PR #1, opened 2019-07-02
Completed Day 2, please review

Ameya145:


MyHashSet.java.txt
```diff
@@ -0,0 +1,31 @@
+class MyHashSet {
+
+    int set[] = new int[1000000];
+    /** Initialize your data structure here. */
+    public MyHashSet() {
+        
+    }
+    
+    public void add(int key) {
+        if(set[key]==0) //if the key is not present, increment index by 1
```
super30admin: Can we add duplicate entries in HashSet?
Think about it.

StackUsingQueues.java.txt
```diff
@@ -0,0 +1,42 @@
+class MyStack {
+
+    Queue<Integer> q = new LinkedList<>();
```
super30admin: It was queue using stacks, not the other way round.

PR #2, opened 2019-07-02
Day 2 Completed:Please Review

mahima95:


Queue using Stack.java
```diff
@@ -0,0 +1,58 @@
...
+            return s2.peek();
+    }
+    return front;
+    }
+    
+    /** Returns whether the queue is empty. */
+    public boolean empty() {
+       return s1.isEmpty() && s2.isEmpty();
+    }
+}
```
super30admin: Well Done !!

MyHashSet.java
```diff
@@ -0,0 +1,44 @@
...
+    public boolean contains(int key) {
+         if(contained[key])
+        { 
+            return true;
+        }
+        
+        return false;
+    }
+}
+
```
super30admin: This is more like a brute force solution.
Try reading more stuff about how HashSet works in java.

PR #5, opened 2019-07-02
Please review my Day 2: Queue using two stacks implementation

MahiteeS:


QueueUsingStack.java
```diff
@@ -0,0 +1,71 @@
...
+    public QueueUsingStack() {
+       stack1 = new Stack<>();
+        stack2 = new Stack<>();
+    }
+    
+    /** Push element x to the back of queue. */
+    
+    // if queue stack is empty, push value in it.
+    // otherwise, transfer the values from queue stack (until its not empty) to reference stack, push new value 
+    // in queue stack and transfer values from reference stack to queue stack.
```
super30admin: Your explanation to this is awesome!!

QueueUsingStack2.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+    
+    /** Push element x to the back of queue. */
+    public void push(int x) {
+          stack1.push(x);
+    }
+    
+    /** Removes the element from in front of queue and returns that element. */
+    // first we update the queue stack with all the newly pushed elements
+    // and then pop the element from queue stack
```
super30admin: Would you like to comment on time and space complexity for all operations in both the approaches.

HashsetUsingArray.java
```diff
@@ -0,0 +1,46 @@
...
+			set.contains(20);
+			set.add(30);
+			set.contains(10);
+			set.remove(10);
+			set.add(10);
+			boolean result =set.contains(40);
+			System.out.println("Hashset contains key result =:"+result);
+			set.remove(10);
+			
+	}
```
super30admin: This is more of brute force solution using lot of space.

QueueUsingStack.java
MahiteeS: O(n) because as soon as we push any element, we move all the elements in the other stack.

QueueUsingStack2.java
MahiteeS: I am not sure as to what exactly can be the complexity, but it seems to be better solution because we only move elements from a stack, when needed i.e. when a peek function is called. I am confused whether it's still O(1) or O(n).

HashsetUsingArray.java
MahiteeS: I'll surely try other ways like linked list and those methods which were made known to us today.

DesignHashsetUsing2DArray.java
```diff
@@ -0,0 +1,59 @@
...
+	        h.contains(1); 
+	        h.contains(3); 
+	        h.add(2); 
+	        h.contains(2);
+	        h.remove(2); 
+	        h.contains(2);
+	         
+	        
+	       
+	    }
```
super30admin: This one is a lot better.
Great Job Mahitee!!

PR #7, opened 2019-07-03
day2

sarang4488:


HashSet.java
```diff
@@ -0,0 +1,48 @@
...
+		// TODO Auto-generated method stub
+		HashSet obj = new HashSet();
+		 obj.add(1);
+		 boolean param = obj.contains(1);
+		 System.out.println(param);
+		 obj.remove(1);
+		  param = obj.contains(1);
+		  System.out.println(param);
+
+	}
```
super30admin: This seems to be more of a brute force solution.

QueueUsingStack.java
```diff
@@ -0,0 +1,58 @@
...
+	public int pop() {
+		if(myStack.empty()) {
+			return -1;
+		}
+		Stack<Integer> temp = new Stack<>();
+		while(!myStack.empty()) {
+			temp.push(myStack.pop());
+		}
+		int value = temp.pop();
+		while(!temp.empty()) {
```
super30admin: Can you optimize this solution?
You need modify something in pop().

PR #10, opened 2019-07-03
completed Day2 PS - please review.

lavanganji:


QueueUsingStacks.java
```diff
@@ -0,0 +1,78 @@
...
+        s1.push(x);
+
+        // Push everything back to s1
+        while (!s2.isEmpty())
+        {
+            s1.push(s2.pop());
+            //s2.pop();
+        }
+    }
+
```
super30admin: Thats a good solution but you need to try both the solutions.
Now go for optimal one.

HashSetImplementation.java
```diff
@@ -0,0 +1,39 @@
...
+
+
+public class HashSetImplementation{
+    public static void main(String[] args){
+          MyHashSet obj = new MyHashSet();
+          obj.add(key);
+          obj.remove(key);
+          boolean param_3 = obj.contains(key);
+
+
```
super30admin: Thats a good brute force solution.
But we need to know optimal solution as well.

PR #11, opened 2019-07-03
Problem1: Two Stack Implementation for Queue

gauravj09:


problem1.cpp
```diff
@@ -0,0 +1,38 @@
...
+        inputStack.push(x);
+    } 
+
+    bool empty(){
+        return inputStack.empty() && outputStack.empty();
+    }
+
+    int pop(){
+        int top = peek();
+        outputStack.pop();
```
super30admin: Dont you think something is wrong here.

PR #19, opened 2019-08-20
Completed Design-2 problems

pradhan1234:
Please Review


pradhan1234:
thanks for reviewing, will do as suggested :)

PR #20, opened 2019-08-20
completed

PATIL1695:


queues.py
```diff
@@ -0,0 +1,53 @@
+class MyQueue(object):
+
+    def __init__(self):
+        self.queue = []
```
kaurramanpreet: we can do this using 2 queues as well
although this is also write way of doing this!!

PR #23, opened 2019-08-22
Completed Design 2 (Please Review)

PoojaR24:


queueusingstack.py
```diff
@@ -0,0 +1,47 @@
...
+        :rtype: None
+        """
+        self.stack.append(x)
+
+    def pop(self):
+        """
+        Removes the element from in front of queue and returns that element.
+        :rtype: int
+        """
+        return self.stack.pop(0)
```
kaurramanpreet: this will remove the elemnt from the end of queue!!
think again!

queueusingstack.py
```diff
@@ -0,0 +1,47 @@
...
+        
+
+    def peek(self):
+        """
+        Get the front element.
+        :rtype: int
+        """
+        return self.stack[0]
```
kaurramanpreet: is this the peek element or the last element?

PR #24, opened 2019-08-22
completed design 2 - Please review

A-B-N:


Queue.java
```diff
@@ -0,0 +1,55 @@
...
+            }
+            
+        }
+       int a= s2.pop();
+       return a;
+    }
+    
+    /** Get the front element. */
+    public int peek() {
+       if(s2.isEmpty()){
```
kaurramanpreet: what if s2 is empty but not s1?

PR #33, opened 2019-10-09
Completed Design-2 - Please review

sindhu819:


sindhu819:
> can you optimise the hast set any further?
> double hashing

ok Will implement that.

PR #59, opened 2019-11-29
Problems 1 and 2.

keerthitejakonuri:
Queue using stacks, HashSet implementation

### DIFF
--- a/MyQueue.java
+++ b/MyQueue.java
@@ -1,0 +1,104 @@
+// Time Complexity : O(1) for push, O(N) for pop, O(N) for peek, O(1) for empty
+// Space Complexity : O(N)
+// Did this code successfully run on Leetcode :
+// Any problem you faced while coding this :
+
+
+// Your code here along with comments explaining your approach
+class MyQueue1 {
+    Stack<Integer> stack1, stack2;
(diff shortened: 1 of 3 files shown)


MyQueue.java
```diff
@@ -0,0 +1,104 @@
...
+
+
+// Your code here along with comments explaining your approach
+class MyQueue1 {
+    Stack<Integer> stack1, stack2;
+    /** Initialize your data structure here. */
+    public MyQueue() {
+        stack1 = new Stack<Integer>();
+        stack2 = new Stack<Integer>();
+    }
```
neha79091: Take care about your class name and constructor name

MyQueue.java
```diff
@@ -0,0 +1,104 @@
...
+    /** Push element x to the back of queue. */
+    public void push(int x) {
+        while(!stack2.isEmpty()){
+            stack1.push(stack2.pop());
+        }
+        stack1.push(x);
+        while(!stack1.isEmpty()){
+            stack2.push(stack1.pop());
+        }
+    }
```
neha79091: No need of doing all this. you can simply push the element in first stack.

PR #61, opened 2019-12-24
Please review complete

nikrasiya:


232_queue_using_stack.py
```diff
@@ -0,0 +1,56 @@
...
+                self.out_stack.append(self.in_stack.pop())
+        return self.out_stack.pop()
+
+    def peek(self) -> int:
+        """
+        Get the front element.
+        """
+        if self.out_stack:
+            return self.out_stack[-1]
+        elif self.in_stack:
```
kaurramanpreet: what if there is no element even in in_Stack?

nikrasiya: It will return None since functions in python return None by default if there is nothing to return.

kaurramanpreet: okay then fine

kaurramanpreet:
good job keep it up!!

PR #64, opened 2019-12-24
Design-2 Compeleted : Please review

Lohith-Aswathappa:


MyQueue.java
```diff
@@ -0,0 +1,60 @@
+package com.company;
+
+import java.util.Stack;
+
+public class MyQueue {
+    Stack<Integer> stack = new Stack<>();
+    public MyQueue() {
+        int pointer = 0;
+        int pointerIndex = -1;
```
kaurramanpreet: good trick to use a variable for this part

jaspindersingh83: @kaurramanpreet this is not a good solution for stack and hashset. We discussed in the class. 

- In hashset @Lohith-Aswathappa is using a boolean array of size 1 million which is not space efficient. 

- In queue solution, he is reversing every time the pop and peek operations are called in. Making it O(n) operation always, however the class discussed solution is O(1) average time. Kindly got through class video

PR #79, opened 2019-12-31
Design 2 

nikhilm2409:


QueuewithStacks.java
```diff
@@ -0,0 +1,62 @@
+// Time Complexity :O(1)
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : Yes
```
super30admin: We are using O(n) space here

Hashset.java
```diff
@@ -0,0 +1,67 @@
+// Time Complexity : O(log n)
+// Space Complexity : O(1)
```
super30admin: Time is O(1) in Java in hashset

PR #85, opened 2020-02-04
Completed Queue Using Stacks problem -- Please Review

AbhiniveshP:
Hi,

I completed Queue using stacks but did it in linear time and space for pop operation. It passed all Leetcode test cases but I'm not sure whether its optimal solution or not. Time and Space Complexities are included in each function.

Thanks
Abhinivesh

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,2 @@
+# Default ignored files
+/workspace.xml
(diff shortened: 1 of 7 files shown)


AbhiniveshP:
Hi,
I have updated both the files with optimal solutions as discussed in class. Please review.
Thanks,
Abhinivesh

PR #87, opened 2020-02-04
Create Problem1_Queue using stacks.py

megrao:
Queues using stacks - operations with a single stack

I wasn't sure if you expect 2 stacks. Hence, I tried with a single stack. Please review.

### DIFF
--- a/Problem1_Queue using stacks.py
+++ b/Problem1_Queue using stacks.py
@@ -1,0 +1,49 @@
+class MyQueue(object):
+
+    def __init__(self):
+        """
+        Initialize your data structure here.
+        """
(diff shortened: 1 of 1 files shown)


Problem1_Queue using stacks.py
```diff
@@ -0,0 +1,49 @@
...
+        """
+        return self.s1.pop(0)
+
+    def peek(self):
+        """
+        Get the front element.
+        :rtype: int
+        """
+        return self.s1[0]
+
```
super30admin: We can use the stack functions of push, pop, peek only `return self.s1[0]` should not be used

PR #95, opened 2020-02-05
Completed Design 2 - Please Review

sayan1995:


queueStack.py
```diff
@@ -0,0 +1,69 @@
+'''
+Time Complexity :
+For push: O(1)
+For pop: O(n)
+
```
super30admin: O(n) is worst case. Average is O(1). Kindly mention it like this in interviews

PR #97, opened 2020-02-06
please review (both questions added)

sivasaimupparaju:


queue.java
```diff
@@ -0,0 +1,76 @@
+//time complexity is o(n)
+//space complexity is o(n)
```
super30admin: Worst case time complexity is O(n) but average case is O(1). Just to make you aware of it

PR #106, opened 2020-03-11
Completed Design II

arcsoftech:


HashSet.java
```diff
@@ -0,0 +1,51 @@
+// Time Complexity : O(N) for searching and remove. O(1) for insertion as I am adding new elemnet at head.
+// Space Complexity : O(N)
+// Did this code successfully run on Leetcode :
```
devashishnyati: Please run this code on leetcode and fill this space

devashishnyati:
Good job Arihant! Nice work with the time and space complexities!
Try HashSet with double hashing also.

HashSet.java
arcsoftech: I did updated the code and pushed my comment on github. Should I raise another pull request ?

devashishnyati: No, you don't have to create a new PR.

PR #107, opened 2020-03-11
Design-2

vaidyapawan:
705. Implementation of HashSet
232. Implement Queue using Stacks

### DIFF
--- a/MyHashSet.java
+++ b/MyHashSet.java
@@ -1,0 +1,46 @@
+// Time Complexity : add: O(1)
+                     remove: O(1)
+                     Contains: O(1)
+// Space Complexity : O(k) where K is size of the initial array.
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : I have created a huge array to implement this, so I have compromized space for better runtime.
+                                             Can you suggest a better option to limit the space.
+
+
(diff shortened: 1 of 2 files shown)


MyHashSet.java
```diff
@@ -0,0 +1,46 @@
+// Time Complexity : add: O(1)
+                     remove: O(1)
+                     Contains: O(1)
+// Space Complexity : O(k) where K is size of the initial array.
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : I have created a huge array to implement this, so I have compromized space for better runtime.
```
devashishnyati: Try using hashing! You can do linear chaining or double hashing to reduce the space.

MyQueue.java
```diff
@@ -0,0 +1,61 @@
...
+    /** Push element x to the back of queue. */
+    public void push(int x) {
+        if(s1.empty()){
+            front = x;
+        }
+        s1.push(x);
+    }
+    
+    /** Removes the element from in front of queue and returns that element. */
+    public int pop() {
```
devashishnyati: This looks like worst case O(n) operation.

devashishnyati:
Good job Pawan!

PR #109, opened 2020-03-11
Completed Design 2: Please Review

KBangera:
I have added the approach taught in yesterday's class for implementing the Hashmap as well


devashishnyati:
Good job Karthik!
Try to optimize the solutions. You can use double hashing and in queue using stacks you can get the time complexities to push -> O(1) get average O(1) worst-case O(n).

PR #111, opened 2020-03-12
Design-2

AishwaryaSharma27:
Please review


devashishnyati:
Good job Aishwarya!
Please add the 4 things on the top from sample.java file.
Try using some kind of hashing for hashset. Try using linear chaining or double hashing with it.

PR #112, opened 2020-03-12
Added code for Design-2 Implementing Queue using stack and implementing HashSet

RutujaKaushike:


devashishnyati:
Good job Rutuja!
Try to implement hashset with double hashing also!

PR #118, opened 2020-03-14
Design-2

nilpatel111296:


devashishnyati:
Good job Nil!
Instead of using rehash, try using double hashing to optimize the code further.
Also, please include whether the code ran successfully on leetcode and any challenges that you faced with the time and space complexities at the top. Refer sample.java file.

PR #119, opened 2020-03-15
Create Queue.java

Iamshuklayash:


devashishnyati:
No need to create 2 PR's for the same problem.
Just update the code and PR will be updated.

PR #124, opened 2020-04-21
Completed Design-2: Please Review

anikpait:


HashSet.java
```diff
@@ -0,0 +1,45 @@
+// Time Complexity :   add(), remove(), contains()  ->  O(1)
+// Space Complexity :  O(n) I am using HashMap to simulate the behavior
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : None
+
+
+// Your code here along with comments explaining your approach
+class MyHashSet {
+    
+    HashMap<Integer, Integer> index;        // Using HashMap to simulate the Hashset behavior
```
super30admin, line 10: Whenever we are asked to make hashing data structure. The intention of interviewer is to find if you can take a basic data structure and implement hashing on it. 

We shouldn't be implementing hashset based on a hashmap

Queue_Using_Stack.java
```diff
@@ -0,0 +1,71 @@
+// Time Complexity : Push()-> O(1), Pop() -> Amortized Cost = O(1) => The first operation will take O(n). After that we can pop at O(1). 
+//                    We will have few/rare O(n) operations only when Stack 2 is empty.
```
super30admin, line 2: Good

super30admin:
Good Job with problem 2. Keep up the hard work

HashSet.java
anikpait: Does an ArrayList implementation work in this case?

super30admin: Why thinking about ArrayList, (still a higher level DS). We will be making it work in O(1) using a normal array

anikpait: Oh okay. 
So, an ArrayList implementation also, if it checks with contains() function on whether the key exists or not and then perform 'add'/ HashSet put() or 'return the key' / HashSet get(), both the HashSet operations will have O(n) implementation resulting the HashSet to have O(n) implementation as well. Is this Correct?

super30admin: I didn't understand. Please ask question in class

PR #130, opened 2020-04-22
Design-2 is complete, please review

Siyami:


queueUsingStacks.js
```diff
@@ -1,7 +1,50 @@
...
+    data = [];
+    push(num) {
+        this.data.push(num);
+    }
+    /**
+ * Removes the element from in front of queue and returns that element.
+ * @return {number}
+ */
+    pop() {
+        return this.data.shift();
```
super30admin, line 22: Lets never use shift going forward

super30admin: We can make the pop also average O(1) as we discussed in class

super30admin:
Please resubmit the solutions based on class discussion

PR #131, opened 2020-04-22
Completed Design2 -- Please review

VenkateswaraRajuDatla:


iivaishnavii:
Good Job, Raju .. You haven't checked if the container has the object, before deleting .Hence, you get the exception. Take care of these cases as well.

VenkateswaraRajuDatla:
Yes, Got it.

PR #143, opened 2020-04-24
Please review

varshakatariya:


Problem_2
```diff
@@ -0,0 +1,78 @@
+/**
+Design Hash Set
+Time Complexity : 
+add() => O(1)
+remove() => O(1)
+contains => O(1)
+
+Space Complexity :
+add() => O(n) => number of collisions
```
super30admin, line 9: this is incorrect. Still O(1). The collisions handled in double hashing for given data range make sure add function is O(1). which part makes you think its O(n)?

super30admin:
Good. Please review inline comments

PR #154, opened 2020-05-25
Design-2 Queue using stack, Please Review

guravashish:
Design-2 
Queue using stack
Please Review, Thank you!

### DIFF
--- a/QueueAsStack.java
+++ b/QueueAsStack.java
@@ -1,0 +1,68 @@
+import java.util.Stack;
+
+// Time Complexity :O(n)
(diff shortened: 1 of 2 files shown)


HashSet.java
```diff
@@ -0,0 +1,49 @@
+// Time Complexity :O(n)
```
chiragshahi, line 1: Define n as well according to the question

chiragshahi:
Great work with the code and comments. Also try linked list solution of set
See inline comment

PR #155, opened 2020-05-25
Completed Design 2. Kindly review and share feedback.

Sumesh-H:


MyQueue.java
```diff
@@ -1,4 +1,14 @@
 import java.util.Stack;
+//Time Complexity : 
+//Push O(1)
+//Pop O(n)
+//Peek O(n)
+//Space Complexity : O(1)
```
chiragshahi, line 6: O(n) where n is the size of the queue

MyHashSet.java
```diff
@@ -0,0 +1,81 @@
+// Time Complexity : add , remove, contains : O(1), O(n) worst case
+// Space Complexity : O(N)
```
chiragshahi, line 2: define n

chiragshahi:
Great work Sumesh especially on hash set question! Be careful of time space complexities 
See inline comment

Sumesh-H:
> Great work Sumesh especially on the hash set question! Be careful of time-space complexities
> See inline comment

Thank you, Chirag. I will work more on understanding time & space complexities

PR #159, opened 2020-05-26
Design 2 completed

AbhijeetDhillon:


AbhijeetDhillon:
Could you please help me with some explanation regarding the time and space complexity for both the problems?

MyHashSet.java
```diff
@@ -0,0 +1,41 @@
+// Time Complexity :O(n)
```
chiragshahi, line 1: this is correct
define n as well 
here n is the length of the array list you're using

MyHashSet.java
```diff
@@ -0,0 +1,41 @@
+// Time Complexity :O(n)
+// Space Complexity :O(1)
```
chiragshahi, line 2: This is incorrect. Here you're using an array list which can n in length, so the space complexity would be O(n)

MyQueue.java
```diff
@@ -0,0 +1,119 @@
+// Time Complexity :O(n)
```
chiragshahi, line 1: it is different for different functions
push, isEmpty O(1)
pop, peek O(n) since you have for loops in those functions

MyQueue.java
```diff
@@ -0,0 +1,119 @@
+// Time Complexity :O(n)
+// Space Complexity :O(1)
```
chiragshahi, line 2: This is incorrect
It is O(n) where n is the size of queue which in turn will be the size of the stack

MyQueue.java
```diff
@@ -0,0 +1,119 @@
...
+        else
+        {
+            return -1;
+        }
+
+    }
+    
+    /** Returns whether the queue is empty. */
+    public boolean empty() {
+        if( stack.empty())
```
chiragshahi: the function is isEmpty()
Is this code working on leetcode?

chiragshahi:
I have added my comments for all the time space complexities you had doubts in
Please be careful when coding and only write yes if it works on leetcode

PR #164, opened 2020-05-26
Design-2 Please review it

nareshkumar-ss:


Sample.java
```diff
@@ -1,7 +1,106 @@
...
+
+
+
+
+
+//Problem 2: Design Hashset
+
+// Time Complexity : Each Operation would take around O(1) as it just access the particular index to retrive the element
+// Space Complexity : Size of the ArrayList which is O(1000000)
+// Did this code successfully run on Leetcode : No, Few scenarios were failing on larger input
```
AbhiniveshP: Instead of creating a 1-D array, try creating a 2-D square matrix and access the element using row and column indices.

AbhiniveshP:
Kindly include a 3-step approach at the beginning describing briefly what your approach was, apart from Time and Space Complexities mentioned.

PR #165, opened 2020-05-26
Design-2 Completed- Please Review

DineshNadar95:


MyHashSet.java
```diff
@@ -0,0 +1,37 @@
+// Time Complexity :O(1)
+// Space Complexity :O(n)
```
chiragshahi, line 2: define n

chiragshahi:
Great job! Make sure to write time complexities for all the functions if they are different from each other 
Also try LL solution of hash set

PR #168, opened 2020-05-26
Design 2 completed - please review

KrishSJ:


Queue.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+	public static void main(String[] args) {
+		Queue myQueue = new Queue();
+		myQueue.push(10);
+		myQueue.push(20);
+		System.out.println(myQueue.pop());
+	}
+}
+//Time Complexity : push - O(n), pop,peek - O(1)
```
AbhiniveshP, line 49: Please check time complexities as the upper bound is not O(1) and O(n) instead

AbhiniveshP:
Well done. Kindly include a 3-step approach at the beginning describing briefly what your approach was, apart from the Time and Space Complexities mentioned.

PR #174, opened 2020-05-26
Completed Design-2 in Java Please Review

sridhar-anish-001430789:


MyHashSet.java
```diff
@@ -0,0 +1,33 @@
...
+
+class MyHashSet {
+
+	/** Initialize your data structure here. */
+	// Declaring an array
+	int set[];
+
+	public MyHashSet() {
+		// Initializing an array
+		set = new int[1000001];
```
Lohith-Aswathappa, line 16: You have used a fixed amount of space, so the space complexity would be O(1)

Lohith-Aswathappa:
Good job, Anish!
Nicely commented.

PR #184, opened 2020-05-27
Created HashSet.py

Shashank001122:
Same issue as HashMap, used Double hashing running very slow in LeetCode

### DIFF
--- a/HashSet.py
+++ b/HashSet.py
@@ -1,0 +1,42 @@
+class MyHashSet:
+
+    def __init__(self):
+        """
+        Initialize your data structure here.
+        """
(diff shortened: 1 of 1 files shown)


HashSet.py
```diff
@@ -0,0 +1,42 @@
+class MyHashSet:
```
Lohith-Aswathappa, line 1: Try the solution taught in class, it'll improve your time as well as space efficiently and also, try to write functions if you feel some part of the code is repeated. In this case, finding the index (key%1000)

Lohith-Aswathappa:
Nice try, Shashank!
Please retry the HashMap solution and review the inline comments

PR #206, opened 2020-07-08
Completed Design_2 - Queue using Stack, HashSet. Please Review it

sneharnair:


Problem-2_Hashset
```diff
@@ -0,0 +1,60 @@
+# Time Complexity : O(1) - get_main_ind, get_aux_ind, add, remove, contains
+# Space Complexity : O(nm) - n : length of the primary array and m : length of secondary array (present at those entries of the primary array which has an entry).
```
samarthjj, line 2: since m,n are 1000, the SC is O(1,000,000) which is O(1)

samarthjj:
Good work Sneha, Please review the Inline comment.
Please push .py files instead of text files :)

PR #212, opened 2020-07-08
Please Review - Design 2 

akashkshekarappa:


MyHashSet.java
```diff
@@ -0,0 +1,61 @@
+// Time Complexity : O(1)
+// Space Complexity : O(n)
```
samarthjj, line 2: SC is O(1000^2) which is a constant, hence O(1).

samarthjj:
Nice work Akash!, Please review the inline comment.

PR #213, opened 2020-07-08
Please review

SaikumarReddySandannagari:


Hash_set.py
```diff
@@ -0,0 +1,53 @@
+class Node:
+    def __init__(self, value, nextNode=None):
+        self.value = value
+        self.next = nextNode
+        
+class Bucket:
+    def __init__(self):
+        self.head = Node(0)
+
+    def insert(self, newValue):
```
AbhijeetDhillon, line 10: what if the newValue has an index that is already present in the hashmap?

AbhijeetDhillon:
Good Job Saikumar. Please mention the time and space complexities along with your approach in the future solutions.

AbhijeetDhillon:
Good Job Saikumar. Please mention the time and space complexities along with your approach in the future solutions.

PR #216, opened 2020-07-08
Implement Queue using Stack - Design 2 - Please review

firoz1905:
I got an error "List out of index during the pop". Kindly give your feedback on that.

### DIFF
--- a/Queue Implementation using Stack
+++ b/Queue Implementation using Stack
@@ -1,0 +1,52 @@
+class MyQueue:
+
+    def __init__(self):
+        """
+        Initialize your data structure here.
+        """
+        self.stack1= []
+        self.stack2= []
+        
+
+    def push(self, x: int) -> None:
+        """
+        Push element x to the back of queue.
+        """
+        self.stack1.append(x)
+        
+        
+
+    def pop(self) -> int:
+        """
+        Removes the element from in front of queue and returns that element.
+        """
+        if self.stack2==None:
+            while self.stack1 !=None:
+                self.stack2.append(self.stack1.pop())
+        return self.stack2.pop()
+        
+
+    def peek(self) -> int:
+        """
+        Get the front element.
+        """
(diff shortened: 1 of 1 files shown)


Queue Implementation using Stack
```diff
@@ -0,0 +1,52 @@
...
+        """
+        self.stack1.append(x)
+        
+        
+
+    def pop(self) -> int:
+        """
+        Removes the element from in front of queue and returns that element.
+        """
+        if self.stack2==None:
```
Sumesh-H, line 23: Where are you checking the out of bound condition?
It should be:
if len(self.stack2) != 0:
            return self.stack2.pop()
        else:
            while (len(self.stack1) != 0):
                self.stack2.append(self.stack1.pop())
        return self.stack2.pop()

Sumesh-H:
Good job! Kindly mention Time(each operation) & Space complexity. Check inline comments.

Queue Implementation using Stack
Sumesh-H, line 23: None is a data type used for defining objects. It is not zero or false. We can assign None to a variable. Use [] or 0 for the code to run. 
i.e.  if self.stack2==[]:
            while self.stack1 !=[]:
                self.stack2.append(self.stack1.pop())
        return self.stack2.pop()

Queue Implementation using Stack
```diff
@@ -0,0 +1,52 @@
...
+            while self.stack1 !=None:
+                self.stack2.append(self.stack1.pop())
+        return self.stack2.pop()
+        
+
+    def peek(self) -> int:
+        """
+        Get the front element.
+        """
+        if self.stack2 == None:
```
Sumesh-H, line 33: if self.stack2 == []:
            while self.stack1!=[]:
                self.stack2.append(self.stack1.pop())       
        return self.stack2[-1]

PR #217, opened 2020-07-08
Design-2 completed in java - please review

Shilpi10:


MyHashSet.java
```diff
@@ -0,0 +1,58 @@
+// Time Complexity : O(1) amortized for each operation
+// Space Complexity : )(n) where n is max size of hashset
```
samarthjj, line 2: Since the size here is a constant i.e 1000. The Space will be O(1000^2), which is O(1).

QueueUsingStack.java
```diff
@@ -0,0 +1,49 @@
+// Time Complexity : O(1) for every operation, amortized for push
```
samarthjj, line 1: Amortized would be when O(n) operations occurs once and then we get O(1) operation cost for the next **n** calls.
In this case for 10 back to back push operations, we only have O(1) for one call and the rest 9 calls are O(n).
Please review the Time Complexity again for push.

samarthjj:
Nice work Shilpi, please review the inline comments.

PR #219, opened 2020-07-08
Completed Design-2 - Please review.

vaibhav-gupta1911:


Problem1.java
```diff
@@ -0,0 +1,78 @@
...
+// Push - Best/Amotized/Worst:O(1)
+// Pop - Best/Amotized:O(1); Worst: O(N)
+// Peek - Best/Amotized:O(1); Worst: O(N)
+
+
+// Space Complexity :
+// O(N), where N is the number of elements
+
+// Did this code successfully run on Leetcode : YES.
+// Any problem you faced while coding this : NO, However, please let me know wheather I am calculating time complexity in the right way.
```
samarthjj, line 21: You are correct with the time complexities. Pop and peek have amortized costs, but push doesn't. Only mention amortized when applicable.

Problem2.java
```diff
@@ -0,0 +1,86 @@
+//PROBLEM - Implement Hashset
+/** 3 Pointer Approcach: 
+ * 2d Arrrsy Boolean array.
+ * 1D array (Bucket) will be indexed [1, 1K] for Val%1000
+ * Each bucketItem array will be indexed for [1, 1K] from Val/1000
+ * We wiil initialize bucketItem arrays on the fly when needed and not in start as they may not be used.
+ * As it is given the numbers are in bound [1, 1M] so 1000 * 1000 will cover every number with NO collisions.
+ * Why Boolean array? - beacause we just need to check if specific value is available in the HashSet boolean True means value present else not.
+ * Approach - On Add item
+ * key%1000 - will give the bucket-Index; key/1000 - will give the bucketItem-Index;
```
samarthjj, line 10: The approach should be a gist and not a step by step explanation. Please write full sentences to describe your approach.

Problem2.java
```diff
@@ -0,0 +1,86 @@
...
+ */ 
+//
+//time Complexity :
+// add - Best/Amotized/Worst:O(1)
+// remove - Best/Amotized:O(1);
+// contains - Best/Amotized:O(1);
+
+
+// Space Complexity :
+// O(Buckets * BucketItems).
```
samarthjj, line 20: Since **Buckets** and **BucketItems** are both 1000 in this case. The TC will be O(1), because O(1Million) is considered a constant.

samarthjj:
Nice work Vaibhav, Please review the inline comments.

PR #220, opened 2020-07-09
Design 2 - Python

Balaviknesh:


HashSet.py
```diff
@@ -0,0 +1,45 @@
+# Time Complexity : getHash1() - O(1), getHash2() - O(1), add()- O(1), remove - O(1), contains - O(1)
+# Space Complexity : o(n^2)
+# Did this code successfully run on Leetcode : yes
+# Any problem you faced while coding this : I see it takes more space, can you suggest what else I can do to reduce it.
```
Sumesh-H, line 4: Bitset instead of a boolean array to save on memory.  This question is for you to understand how HashSet works. To understand the trade-offs of a hash function and how well you are handling collision.

Sumesh-H:
Good work Balaviknesh! Try the other problem as well. Check inline comment.

Sumesh-H:
Good work Balaviknesh! Try the other problem as well. Check inline comment.

PR #234, opened 2020-07-11
completed. Please review.

satishsolanki1990:
Hi, Sorry for the delay.

### DIFF
--- a/Queue_using_stacks.py
+++ b/Queue_using_stacks.py
@@ -1,0 +1,83 @@
+class stack:
+    def __init__(self):
+        self.a=[]
+    def push(self,a):
+        self.a.append(a)
+
+    def pop(self):
+        return self.a.pop()
+
+    def peek(self):
+        if not self.isEmpty():
+            return self.a[-1]
+
+    def isEmpty(self):
+        if len(self.a) == 0:
+            return True
+        else:
+            return False
+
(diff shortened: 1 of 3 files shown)


Queue_using_stacks.py
```diff
@@ -0,0 +1,83 @@
+class stack:
+    def __init__(self):
+        self.a=[]
+    def push(self,a):
+        self.a.append(a)
+
+    def pop(self):
+        return self.a.pop()
```
samarthjj, line 8: What if pop is called on an empty stack, please add edge cases to error guard

Sample.java
```diff
@@ -1,7 +1,15 @@
-// Time Complexity :
-// Space Complexity :
-// Did this code successfully run on Leetcode :
-// Any problem you faced while coding this :
+//  problem 1: implement-queue-using-stacks/
+// Time Complexity : 32 ms, faster than 52.14%
+// Space Complexity : 13.9 MB, less than 49.36%
```
samarthjj, line 3: Time & Space mentioned has to be asymptotic not stats from leetcode.

samarthjj:
Good work Satish, review the inline comments.
For Time & Space please copy and paste the template to the same files as the problem instead of in sample.java.
Please add comments explaining your approach as well.

PR #244, opened 2020-08-17
Design-2 problems-implementation

Jayesh-sathe:


Gupta-Keshav-001378461:
Your Queue using stack code is not working on Leetcode.
And the Hashset solution is incorrect because it is mentioned You should not use built in libraries.

Jayesh-sathe:
Please check the updated code after correction.
Thank you.

Gupta-Keshav-001378461:
Good Job!!

PR #248, opened 2020-08-17
Completed Design 2 problems - Please review 

rimzimt:


hashset.py
```diff
@@ -0,0 +1,61 @@
...
+        
+    def getHashValue(self,key):
+        
+        return key%self.k
+
+    def add(self, key: int) -> None:
+        
+        hashVal=self.getHashValue(key)
+        
+        if key not in self.hash[hashVal]:
```
samarthjj, line 27: O(n) operation, linearly scans the entire array to check if key exists

hashset.py
```diff
@@ -0,0 +1,61 @@
+# Time Complexity : O(1)
```
samarthjj, line 1: O(n^2). All keys are stored at a single index and remove triggers shifting.

queue.py
```diff
@@ -0,0 +1,61 @@
...
+        """
+        self.stack.append(x)
+        
+
+    def pop(self) -> int:
+        """
+        Removes the element from in front of queue and returns that element.
+        """
+        
+        return self.stack.pop(0)
```
samarthjj, line 33: removes from index 0, shifts all n-1 elements. O(n) operation.

queue.py
```diff
@@ -0,0 +1,61 @@
+# Time Complexity : O(1)
```
samarthjj, line 1: O(n)

samarthjj:
Good work Rimzim, Please review the inline comments. 
Please also implement both the problems the way they were done in class.

samarthjj:
Good work Rimzim, Please review the inline comments. 
Please also implement both the problems the way they were done in class.

rimzimt:
Updated with class approach. Please review.

PR #255, opened 2020-08-18
Completed Design-2- Please Review

sahanaramesh09:


MyHashSet.java
```diff
@@ -0,0 +1,57 @@
+// Time Complexity : o(1)
+// Space Complexity : o(1000001)
```
samarthjj, line 2: Constants are ignore, The space here should be O(1)

samarthjj:
Good work Sahana, Please review the inline comments. 
Please also implement The HashSet problem the way it was done in class.

samarthjj:
Good work Sahana, Please review the inline comments. 
Please also implement The HashSet problem the way it was done in class.

PR #257, opened 2020-08-18
Design-2, Please review

rohit-gorle:
I am not getting why my 2nd problem(HashSet) couldn't pass all test cases. please correct me where I went wrong. and let me know if my time and space complexities are right.


Gupta-Keshav-001378461:
Good Job!!
Try the question with the chaining. rather than linked list.

PR #261, opened 2020-08-18
Completed Day2 Problems. Please review and give feedback especially regarding HashSet implementation.

ShaivyT:


Sample.java
```diff
@@ -1,7 +1,71 @@
-// Time Complexity :
-// Space Complexity :
-// Did this code successfully run on Leetcode :
-// Any problem you faced while coding this :
-
+// Time Complexity : All operations except push - O(1), push operation - O(n)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : I didn't think of maintaining a variable to keep track of the top element. 
+// I wrote it in the second version. Can you please mention in the comments later that how should we think of this from the beginning?
```
samarthjj, line 5: These little developments such as maintaining a front pointer would become more intuitive as we proceed along the course. Please make sure to whiteboard your problem really well, helps in arriving to a better solution.

Sample2.java
```diff
@@ -0,0 +1,56 @@
...
+        if (!arr[hash].contains(key)) { 
+            arr[hash].add(key);
+        }
+        
+    }
+    
+    public void remove(int key) {
+        int hash = generateHash(key);        
+                
+        if (arr[hash].contains(key)) {
```
samarthjj, line 38: O(n^2) operation. Linearly searches and then removing causes shifting as well.

samarthjj:
Good work Shaivy, Please review the inline comments. 
Please also implement The HashSet problem the way it was done in class.

samarthjj:
Good work Shaivy, Please review the inline comments. 
Please also implement The HashSet problem the way it was done in class.

Sample2.java
ShaivyT: Shouldn't it be O(N * K)? Considering different sizes of the array and arraylist?

PR #262, opened 2020-08-18
Completed Design-2; Queue and HashSet

rusty-sj:


keshav-Gupta-IS (commented):
Design Hashset code is not working on leetcode. It will give you the error of Time Limit Exceeded.
Good Job with the queue using stack!!

Gupta-Keshav-001378461:
Design Hashset code is not working on leetcode. It will give you the error of Time Limit Exceeded.
Good Job with the queue using stack!!

rusty-sj:
Reduced time from ~1565 ms to 23 ms. Oddly, the previous solution was being accepted too, but nonetheless, I fixed it now. Please review.

PR #387, opened 2020-12-21
Completed Sprint Design-2: Please review

shashwathkamath:


MyHashSet.java
```diff
@@ -0,0 +1,70 @@
...
+        return key%MAX;
+    }
+
+    private int getPos(int key,int index){
+
+        if (set[index] == null){
+            return -1;
+        }
+        else{
+            for (int i = 0; i < set[index].size(); i++){
```
samarthjj, line 26: Not a constant time operation.

MyHashSet.java
```diff
@@ -0,0 +1,70 @@
+// Time Complexity : O(1) to add, udpdate
```
samarthjj, line 1: If all elements are stored in a single bucket, the time complexity here will be O(n).

samarthjj:
Good work here Shashwath, Please review the inline comments.
Please also add a brief description of your approach in **3-4 complete English Sentences, not pseudocode**.

samarthjj:
Good work here Shashwath, Please review the inline comments.
Please also add a brief description of your approach in **3-4 complete English Sentences, not pseudocode**.

HashTable.java
```diff
@@ -0,0 +1,107 @@
...
+            if (numOfSlots % i == 0){
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private int getNextPrimeNumber(int numOfSlots){
+        int i = numOfSlots;
+        while (true){
```
samarthjj, line 37: A better way to use boolean logic would be **"while(! isPrime( i ))"**

HashTable.java
```diff
@@ -0,0 +1,107 @@
...
+    element in new position
+     */
+
+    private int hashFunction1(String word){
+        int index = word.hashCode();
+        index = index % arrSize;
+
+        //sometimes Java can return negative hashcode, hence  we will try to avoid this
+        if (index < 0){
+            index += arrSize;
```
samarthjj, line 58: What is the guarantee that by incrementing index by arrSize once it will become a positive integer?

HashTable.java
```diff
@@ -0,0 +1,107 @@
+class HashTable{
```
samarthjj, line 1: Time & Space?

HashTable.java
kamathshashwath1991: okay..that makes sense.

HashTable.java
kamathshashwath1991: Time in case of insert should be O(1) but in case of search worst case will be still O(n) but I feel it will perform better as we are utilizing hashfunctions even to search the index and then to check whether the element is present or not.
Space complexity would be O(N) as we are using one String array to store the element.,

PR #440, opened 2021-01-01
Design-2 

SwapnilDhavale:


Problem-1.py
```diff
@@ -0,0 +1,61 @@
...
+#Did this code successfully run on Leetcode :yes
+#Any problem you faced while coding this :no
+class MyQueue(object):
+
+    def __init__(self):
+        """
+        Initialize your data structure here.
+        """
+        self.inStack=deque()
+        self.outStack=deque()
```
samarthjj, line 13: While the entire logic is correct, it would be better to use a stack since the question explicitly asks for it.

samarthjj:
Good job Swapnil, Please review the inline comments.

samarthjj:
Good job Swapnil, Please review the inline comments.

PR #473, opened 2021-02-16
Please review

ShilpaCatherine:


Sample.java
```diff
@@ -1,7 +1,88 @@
...
+            loan();
+            stack1.push(x);
+        }else {
+            stack1.push(x);
+        }
+    }
+
+    public void loan(){
+        if(stack1.size() == 0 && stack2.size() !=0) {
+            for(int i = 0 ; i < stack2.size() ; i++ ){
```
Lohith-Aswathappa, line 33: Consider the size of stack 2 is 6
if you use for loop, you only pop out 3 elements. 
Every time you pop, stack2 size decreases and i increases. That's the reason why you won't be able to get the minimum elements.

Lohith-Aswathappa (commented):
Nice try, Shilpa!
Check inline comments, try the approach as taught in class

Lohith-Aswathappa:
Nice try, Shilpa!
Check inline comments, try the approach as taught in class

PR #489, opened 2021-02-17
Design 2, Problem 1:Q Using Stack

sachin0112358:


QUsingStack.java
```diff
@@ -0,0 +1,61 @@
+class MyQueue {
+
+    /** Initialize your data structure here. */
+    Stack<Integer> lifo;
```
yahska-ednihs, line 4: Please add the other problem. Also please add comments explaining your approach and time and space complexities.

PR #502, opened 2021-02-25
done in java

stuti19989:


khushal1198 (commented):
Good. Please update with problems with optimal solutions.

khushal1198:
Good. Please update with problems with optimal solutions.

PR #514, opened 2021-03-17

kunika94:


MyHashSet.java
```diff
@@ -0,0 +1,54 @@
...
+    /** Initialize your data structure here. */
+    int Bucket; //to calcute the bucket number
+    int BucketItems; // to calculate the index of the item number in a given bucket
+    boolean[][] storage;
+    public MyHashSet() {
+        Bucket = 1000;
+        BucketItems = 1001;
+        storage = new boolean[Bucket][];
+    }
+    private int Bucket(int key){
```
rubieHess, line 12: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

kunika94: Thanks for pointing out the naming convention techniques and I'll make sure that I follow them. Is there anything wrong with the code in terms of functionality as well?

rubieHess: Happy to help. Nothing else.

khushal1198 (commented):
Good job with the code and comments!

khushal1198:
Good job with the code and comments!

PR #529, opened 2021-03-21
Solutions for Design2

dp2909:
1) Is there a way in Python where I can declare global variables and use them in all the methods. For example in the Hashset implementation I had to use the buckets and bucketitems variables. I had to define them as self.bucketitems and self.buckets in the init method. Can i do it somewhere else? So that i can use them as just buckets and bucketitmes rather than with the self keyword.
2) I also had to do the same with the additional methods I had to create such as locate_buckets as object methods and had to pass the object itself as the parameter. I had to use them inside the other methods as self.method name. Can I somehow define them without the self keyword?

### DIFF
--- a/Queue.py
+++ b/Queue.py
@@ -1,0 +1,56 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# In[1]:
+
+
+class MyQueue:
+
+    def __init__(self):
+        """
+        Initialize your data structure here.
+        """
+        self.instack = []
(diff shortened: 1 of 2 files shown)


yahska-ednihs:
You can use global, local and nonlocal keywords but personally I prefer the self.___ approach. You can use Classname.method() for static methods.

yahska-ednihs:
Good Job! Add comments and time and space complexities for all the operations

PR #597, opened 2021-05-26
added hashset and queue using stack

yangd01234:
Please review my code thanks


maithrreye:
Good Job ! Please add TC and SC.

maithrreye:
Good Job ! Please add TC and SC.

PR #606, opened 2021-05-27
Done with Design 2 - Please review

BhavyaSri1001:


BhavyaSri1001:
Also I have changed the code for hashmap Implementation which is present in Design 1 as i have submitted it previously with O(n) complexity. Please review that one as well.

PR #623, opened 2021-05-27
complete

shivamarr96:


Gupta-Keshav-001378461 (commented):
Good Job, Shivam!! Keep it up. One suggestion I would give is to write comments in your code it will help you when you revisit these codes again.

Gupta-Keshav-001378461:
Good Job, Shivam!! Keep it up. One suggestion I would give is to write comments in your code it will help you when you revisit these codes again.

PR #631, opened 2021-05-27
Solved design2 exercises

suman123rock:
please review


khushal1198 (commented):
Good work with code. Please add complexities too.

khushal1198:
Good work with code. Please add complexities too.

PR #645, opened 2021-06-01
Design 2 Completed

NishigandhaMhaisne:


hashset.py
```diff
@@ -0,0 +1,41 @@
+# Time Complexity: O(1)
+# Space Complexity : O(1)
+
```
Lohith-Aswathappa, line 3: try the approach taught in class as well

Lohith-Aswathappa (commented):
Well done, Nishigandha!
check inline comments

Lohith-Aswathappa:
Well done, Nishigandha!
check inline comments

PR #648, opened 2021-06-01
HashSet + Queue

hharshitarora:
For the queue solution I am still not sure about the amortized solution I believe I did O(n) for every operation.
For hashset I used the 2D Array method, its just that I have a very high runtime on leetcode even though I believe I did all operations in O(1). Please correct me if I am wrong.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,3 @@
+# Default ignored files
+/shelf/
+/workspace.xml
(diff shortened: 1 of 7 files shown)


Rooppesh:
Good Job! The runtime on leetcode doesn't matter. The amortized time complexity for Hashset solution would be O(1). The amotized time complexity for Queue will also be O(1).

Rooppesh:
Good Job! The runtime on leetcode doesn't matter. The amortized time complexity for Hashset solution would be O(1). The amotized time complexity for Queue will also be O(1).

PR #652, opened 2021-06-01
[Design-2][Java] Please review

mahesh-9:


Rooppesh:
Good Job Mahesh! The amortized Time Complexity for Pop and Peek in Queue would be O(1). Please check.

Rooppesh:
Good Job Mahesh! The amortized Time Complexity for Pop and Peek in Queue would be O(1). Please check.

PR #704, opened 2021-07-21
Design 2

SaiSruthiMareedu:


khushal1198 (commented):
Good job with both the problems Sruthi!

khushal1198:
Good job with both the problems Sruthi!

PR #710, opened 2021-07-21
Completed hashset in python file . had doubts in second question

psambalkar:


Rooppesh:
Good Job Parth! Let me know if there are any specific doubts!

Rooppesh:
Good Job Parth! Let me know if there are any specific doubts!

PR #737, opened 2021-08-18
Design-2 completed, Please review.

shilpagowda812:


sridhar-anish (commented):
Good job with everything, make sure to write the TC and SC  as well

sridhar-anish:
Good job with everything, make sure to write the TC and SC  as well

PR #743, opened 2021-08-18
Updated code for designing hashset, 4tc's passing over leetc0de-showing excpetion for add/containing/removing value 1000000, please review"

veeceey:
Output: [null,null,null,true,false,null,true,null,true]
Expected: [null,null,null,true,false,null,true,null,false]


veeceey:
"Updated code for hashset, 4tc's passing over leetc0de-showing excpetion for add/containing/removing value 1000000, please review"

PR #754, opened 2021-08-19
Update Sample.java

hanut1909:


manojvarmarudraraju (commented):
Good job on question 1, but amortized time complexity for the pop operation will O(1) please check, complete the other question as well

manojvarmarudraraju:
Good job on question 1, but amortized time complexity for the pop operation will O(1) please check, complete the other question as well

PR #795, opened 2021-09-14
completed problem 2

parthp2:


Stnagam:
Please Add TC SC

Stnagam:
Please Add TC SC

PR #807, opened 2021-09-15
completed Design -2 Hashset- Please Review

niharikagupta2021:


hashSet.py
```diff
@@ -0,0 +1,60 @@
+"""
```
manojvarmarudraraju, line 1: Try using a boolean array of 1000*1000 it will reduce the time and space complexities of remove functions

manojvarmarudraraju (commented):
Good job, please complete the other problem as well

manojvarmarudraraju:
Good job, please complete the other problem as well

PR #874, opened 2021-11-25
Completed Design-2. Having an issue with hashmap implementation.

SaiGowtham-11:
I have implemented Hashmap implementation but it only was able to pass 27/36 cases. I tried to debug but unable to beacuse the input in the test case is too big. Please let me know if you find any issue with my code.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -1,0 +1,51 @@
+#include <stdio.h>
+#include "stack"
+using namespace std;
+
+class MyQueue {
+public:
+    stack<int> s1;//1,2,
+    stack<int> s2;
+    MyQueue() {
+
+    }
+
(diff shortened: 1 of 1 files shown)


main.cpp
```diff
@@ -0,0 +1,57 @@
+#include <stdio.h>
```
manojvarmarudraraju, line 1: first bucket should accomodate one more element because 1000000 will be coming in at 1000th index

manojvarmarudraraju: So add a condition for that case

manojvarmarudraraju (commented):
Good job, please try the other problem as well

manojvarmarudraraju:
Good job, please try the other problem as well

PR #901, opened 2021-12-21
S30 FAANMG Problem #2

ehassah10:
Design a SplStack that supports all the stack operations push, pop, isEmpty, isFull and an operation getMin() which should return minimum element from the SplStack. All these operations must be O(1). To implement this, you should only use standard Stack data structure.

### DIFF
--- a/Sample.java
+++ b/Sample.java
@@ -1,7 +1,108 @@
-// Time Complexity :
-// Space Complexity :
-// Did this code successfully run on Leetcode :
-// Any problem you faced while coding this :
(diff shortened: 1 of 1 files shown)


brunda4 (commented):
Please provide both problems of Design-01

brunda4:
Please provide both problems of Design-01

ehassah10:
> Please provide both problems of Design-01
added in this PR https://github.com/super30admin/Design-1/pull/944

PR #903, opened 2021-12-21
Completed Problem 2 from Design 2

anmols97:
I implemented the solution using boolean array. I tried to implement the solution using list but was unable to do so as I am inept at coding complex problems. Nonetheless, I tried to understand the implementation conceptually and was able to get a good picture of the solution.


Problem2-HashSet.java 
```diff
@@ -0,0 +1,32 @@
+// Time Complexity:O(1)
```
manojvarmarudraraju, line 1: Please try double hasing

manojvarmarudraraju (commented):
Good job!!

manojvarmarudraraju:
Good job!!

PR #904, opened 2021-12-21
Completed design-2 please review

Mohit1215:


Mohit1215:
I tried implementing double hashing as we discussed in yesterday's class. I tried coding the max but failed the huge test case of the big test case. I was able to clear 26/32 test cases for hash set and all 21/21 test cases for queue using stacks.

sonalgupta276 (commented):
Good job!

sonalgupta276:
Good job!

PR #954, opened 2021-12-28
python solution for design 2

gdg4dev:


sonalgupta276 (commented):
Mention TC and SC

sonalgupta276:
Mention TC and SC

gdg4dev:
please check again, i've updated the files :)

PR #963, opened 2022-01-31
Design-2 implement HashSet Please Review

yash9522:


HashSet.cpp
```diff
@@ -0,0 +1,38 @@
...
+ 3. To remove an element, I am making that bit as 0.
+ 4. hashSet Contains method checks whether the bit for the given key is set or not and return the value accoring to it.
+  */
+
+
+
+class MyHashSet {
+public:
+    bitset<1000001> b;
+
```
shrivatsav, line 22: Good! But this is similar to having a hash table of size(n) where n is the maximum key value. We need to reduce the space as well ( worse case it'll be O(n)) keeping accessing as O(1).

shrivatsav (commented):
Excellent Comments! Good Work! Please check the feedback.

shrivatsav:
Excellent Comments! Good Work! Please check the feedback.

PR #1009, opened 2022-03-01
Completed Design-2 Problem-2 - Please Review

suyog30:


HashSet.py
```diff
@@ -0,0 +1,64 @@
+# Time Complexity : O(1) for searching and adding
+# Space Complexity : O(1)
+# Did this code successfully run on Leetcode : Yes
+# Any problem you faced while coding this : No
+
+#Used the logic of HashSets in Java and implemented tehir storage using two classes for Buckets and the actual Hashset.
+
+class Bucket:
+    def __init__(self):
```
manojvarmarudraraju, line 9: This will be O(n) solution because u have to search for it, instead make each bucket some fixed size array and based on the second hash add it in the respective space

HashSet.py
```diff
@@ -0,0 +1,64 @@
+# Time Complexity : O(1) for searching and adding
+# Space Complexity : O(1)
```
manojvarmarudraraju, line 2: Space complexity of a hashset is O(n)

manojvarmarudraraju (commented):
Good job, please check the suggestions!!

manojvarmarudraraju:
Good job, please check the suggestions!!

PR #1053, opened 2022-03-30
Completed HashSet using Double Hashing

adityamulik:


MyHashSet.java
```diff
@@ -0,0 +1,79 @@
+
+public class MyHashSet {
+    /*
+     * Time Complexity: O(1) for adding, removing and searching an element in HashSet
+     * Space Complexity: A little sketchy on this. I feel this is O(n) as the fixed size of array for buckets (primary
+     * array is given). Then O(log(n)) for secondary array in the best case. Can this be O(n(log(n)) considering in the
```
shrivatsav, line 6: The Worse case complexity can be states as `O(k^2)` here where k is the size of the bucket.

shrivatsav (commented):
Good Work!

shrivatsav:
Good Work!

PR #1097, opened 2022-04-27
Design HashSet

arshmohd:
Implemented HashSet with following requested format: 

class MyHashSet {

    public MyHashSet() {
        
    }
    
    public void add(int key) {
        
    }
    
    public void remove(int key) {
        
    }
    
    public boolean contains(int key) {
        
    }
}

Also included the following test case: 

    MyHashSet hs = new MyHashSet();
    hs.add(1);
    hs.add(2);
    System.out.println();
    boolean is1Present = hs.contains(1); //true
    System.out.println(is1Present);
    boolean is3Present = hs.contains(3); // false
    System.out.println(is3Present);
    hs.add(4);
    boolean is5Present = hs.contains(5); //false
    System.out.println(is5Present);
    boolean is4Present = hs.contains(4); //true
    System.out.println(is4Present);

### DIFF
--- a/MyHashSet.java
+++ b/MyHashSet.java
@@ -1,0 +1,70 @@
+mport java.util.*;
+
+import java.util.*;
+
+public class MyHashSet {
+    private final int MAX_VAL = 1000000;
+    private final int sizeOfArray = 100;
+    private List<List<Integer>> mainBucket;
+
+    public MyHashSet() {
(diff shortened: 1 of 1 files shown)


meghasn (commented):
Good job on both the problems

meghasn:
Good job on both the problems

PR #1120, opened 2022-04-28
Completed S30 FAANMG Problem #3 {Easy}. Please Review.

SaurabhGujare:
Design the following operations of a queue

push(x) -- Push element x to the back of queue.
pop() -- Removes the element from in front of queue.
peek() -- Get the front element.
empty() -- Return whether the queue is empty.


meghasn (commented):
Good Job!

meghasn:
Good Job!

PR #1145, opened 2022-05-24
Completed Design2

shwetajose:


shwetajose:
Design 2 - Queues as Stack

keshav-Gupta-IS (commented):
You can use stack data structure in both these problems. Also try to put comments explaining the approach.

keshav-Gupta-IS:
You can use stack data structure in both these problems. Also try to put comments explaining the approach.

PR #1200, opened 2022-05-26
Complete Design-2 in python.

tathyapatel:
Complete Design-2 in python.

### DIFF
--- a/HashSet.py
+++ b/HashSet.py
@@ -1,0 +1,74 @@
+"""
+// Time Complexity : O(1) for all methods
+// Space Complexity : O(n + log(n) )
+// Did this code successfully run on Leetcode : Yes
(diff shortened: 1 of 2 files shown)


HashSet.py
```diff
@@ -0,0 +1,74 @@
+"""
+// Time Complexity : O(1) for all methods
+// Space Complexity : O(n + log(n) )
```
Sumesh-H, line 3: why log(n)?

Sumesh-H (commented):
Good Job Tathya, please address the inline comments.

Sumesh-H:
Good Job Tathya, please address the inline comments.

HashSet.py
tathyapatel: Because sub-array is not created for every element in the primary array.

PR #1229, opened 2022-06-22
Design-2 HashSet problem

pranjaliseth:


jeelpatel18 (commented):
Good work .Please add Tc ands SC and also upload the 2nd problem as well.

jeelpatel18:
Okay, Design-1 is having MinStack and HashSet problems and Design-2 is having HashMap and Queue using stacks. So, that's why I have commented.

pranjaliseth:
Oh okay. I was under the idea that the first 2 questions were a part of design-1 and 2, as mentioned in the git link. I'll upload Design-2 soon. Thank you.

PR #1233, opened 2022-06-22
feature(hashset): implement hashset without using built in libraries

mounica-bashaboyina:


HashMap.js
```diff
@@ -0,0 +1,34 @@
...
+  find = (head, key) => {
+    const prev = head;
+    const curr = head.next;
+    while (curr != null && curr.key != key) {
+      prev = curr;
+      curr = curr.next;
+    }
+    return prev;
+  };
+}
```
mounica-bashaboyina, line 34: Facing difficulties while implementing few things in javascript since it is not typed. i will try writing using typescript.

keshav-Gupta-IS (commented):
Good Work, Mounica!! I would suggest to pick up the language you are already comfortable in  using data-structures.

keshav-Gupta-IS:
Good Work, Mounica!! I would suggest to pick up the language you already comfortable in  using data-structures.

PR #1253, opened 2022-06-23
Completed Sprint Design-2

siddbanerjee2012:


Design Hashmap.java
```diff
@@ -0,0 +1,27 @@
+// Time Complexity : O(1)
+// Space Complexity :O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : Yes
```
jeelpatel18, line 4: What problems you faced?

Design Hashmap.java
```diff
@@ -0,0 +1,27 @@
...
+
+
+// Your code here along with comments explaining your approach
```
jeelpatel18, line 7: Implement your hash map by doing hashing on key, you don't have to create array of integers of size 1000001

jeelpatel18: Use hashing on given key for get(), put() and remove() methods.

Implement Queue Using Stacks.java
```diff
@@ -0,0 +1,45 @@
+// Time Complexity :O(1)
+// Space Complexity :O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : Yes
```
jeelpatel18, line 4: What are they?

Design Hashmap.java
jeelpatel18: Do you still have confusion on those problems?

Implement Queue Using Stacks.java
jeelpatel18: Do you still have confusion on those problems?

jeelpatel18 (commented):
Good job....

jeelpatel18:
Good job....

PR #1259, opened 2022-06-23

sahithi6445:


HashMap.java
```diff
@@ -0,0 +1,85 @@
+// Time Complexity : O(1)
```
daredevilteja, line 1: TC should be O(n) because in worst case you would be iterating through each node

HashMap.java
```diff
@@ -0,0 +1,85 @@
...
+// Space Complexity : O(1)
```
daredevilteja, line 2: SC should be O(n) because you are using a linkedlist

daredevilteja (commented):
Good Work! Keep it up

daredevilteja:
Good Work! Keep it up

PR #1317, opened 2022-07-22
completed Design-2

KeerthanaThonupunuri:


queuestack.py
```diff
@@ -0,0 +1,42 @@
+#Implement queue using stacks
+#Executed on leet code
+#Time complexity - push-O(1), Pop - O(1), Peek- O(n)
```
daredevilteja, line 3: Pop - O(n) because you are performing peek operation there also

daredevilteja (commented):
Please check the comments

daredevilteja:
Good Work! Keep it up

PR #1336, opened 2022-08-17
Attempted Before class Sprint Design2 Please Review

BhargavVasudevaVunnam:


daredevilteja:
Good Work! Keep it up

Try to use meaningful variable names and don't use `var` keyword. Instead of `var` keyword use the correct datatype.

PR #1368, opened 2022-08-30
Completed design-2 please review

SnehaNellutla:


MyHashMap.java
```diff
@@ -0,0 +1,74 @@
+// Time Complexity : O(1)
```
daredevilteja, line 1: TC is O(n) for find operation

MyQueue.java
```diff
@@ -0,0 +1,62 @@
+// Time Complexity :O(1) for push, O(1) for avg pop case and worst case O(n), O(1) for peek
```
daredevilteja, line 1: TC is O(n) for peek in worst case

daredevilteja:
Good Work! Keep it up

PR #1375, opened 2022-09-08
Design 2- Sprint Completed

rishabparekh14:
Please check the hash Map approach if this is not good please suggest an alternate approach i.e. can I use dictionary but is it by default a form of HashMap in python

### DIFF
--- a/Problem2.py
+++ b/Problem2.py
@@ -1,0 +1,62 @@
+# Time Complexity :- contains O(1) for put , O(N+1) get , O(N+1) remove
+# # Space Complexity : - O(2N) where N is the number of elements and 2N is because we are using 2 arrays 
+# Did this code successfully run on Leetcode : Yes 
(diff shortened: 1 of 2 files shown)


Problem2.py
```diff
@@ -0,0 +1,62 @@
+# Time Complexity :- contains O(1) for put , O(N+1) get , O(N+1) remove
+# # Space Complexity : - O(2N) where N is the number of elements and 2N is because we are using 2 arrays 
+# Did this code successfully run on Leetcode : Yes 
+# Any problem you faced while coding this : Yes the code did run on leetcode but the performance is not that better thought of alternate solution but still not that good performance
```
daredevilteja, line 4: You can try using double hashing here

Problem2.py
```diff
@@ -0,0 +1,62 @@
+# Time Complexity :- contains O(1) for put , O(N+1) get , O(N+1) remove
```
daredevilteja, line 1: Please recheck the TC for put method

PR #1379, opened 2022-09-08
Design 2 completed.

samritabakshi:


daredevilteja:
Please mention TC, SC and whether it ran on leetcode for every problem

PR #1381, opened 2022-09-08
Completed Sprint Design2 - Please review

anakkah:


queue_with_stacks_in_class.py
```diff
@@ -0,0 +1,32 @@
...
+# Space Complexity: O(1)
+# Did this code successfully run on Leetcode : Yes
+# Any problem you faced while coding this : No
+
+class MyQueue:
+
+    def __init__(self):
+        self.stack1 = []
+        self.stack2 = []
+
```
anakkah, line 11: This code only works when the values to be added to the queue are in the range [1,9].
As the requirements say that x (input value) lies in that range, I opted for this logic
Logic:
Add the first elem to stack1
Add size of stack1 to stack2
While adding second element, get the prev val added(say 3) and add the prev val * 10 + current val (5), so the val to be pushed to the stack becomes 3*10 + 5 = 35
Next value to add is say 4, then the value in stack becomes 35*10 + 4 = 354
While adding the elements, keep track of the size of the stack in the second stack
Pop the elements using // and % and get the first values

I think the complexity is O(1), but not clear on it. Please confirm.

queue_with_stacks.py
```diff
@@ -0,0 +1,46 @@
...
+        self.stack1 = []
+        self.stack2 = [0]
+
+    def push(self, x: int) -> None:
+        size = self.stack2[-1]
+        if size > 0:
+            top = self.stack1[-1]
+            x = top * 10 + x
+        self.stack1.append(x)
+        self.stack2.append(size + 1)
```
anakkah, line 18: This code only works when the values to be added to the queue are in the range [1,9].
As the requirements say that x (input value) lies in that range, I opted for this logic
Logic:
Add the first elem to stack1
Add size of stack1 to stack2
While adding second element, get the prev val added(say 3) and add the prev val * 10 + current val (5), so the val to be pushed to the stack becomes 3 * 10 + 5 = 35
Next value to add is say 4, then the value in stack becomes 35 * 10 + 4 = 354
While adding the elements, keep track of the size of the stack in the second stack
Pop the elements using // and % and get the first values

I think the complexity is O(1), but not clear on it. Please confirm.

meghasn: Please do check the video posted and try to follow that method.

meghasn (commented):
Good Job! Please go through the video once again

meghasn:
Good Job! Please go through the video once again

PR #1397, opened 2022-09-09
Completed Design-2 Please review

sh-ruchi:


HashMap.java
```diff
@@ -0,0 +1,94 @@
+// Time complexity : O(1) for all the operations
```
daredevilteja, line 1: Please recheck the TC

daredevilteja:
Please mention TC, SC and whether it ran successfully on leetcode or not, for every problem

PR #1405, opened 2022-09-10
Completed Design-2 -- Please review

nandishamin0:


implement-queue-using-stacks.py
```diff
@@ -0,0 +1,36 @@
+# Time complexity: O(1)
+# Space complexity: O(1)
```
daredevilteja, line 2: SC is O(n) where n is stack size

daredevilteja:
Good Work! Keep it up

PR #1410, opened 2022-09-12
Completed Design-2 (Please review)

mananshah4397:


daredevilteja:
Please recheck the mentioned TC, SC and also mention whether the problems ran successfully on leetcode

mananshah4397:
Yes, the problems ran successfully on leetcode.
Also, would the time and space complexities in all cases be constant because we have kept the initial array size (10^4) and linked list size (10^2) constant.
Or is it not the case? Could you please explain me, I'm confused now!

mananshah4397:
The above message was in context of HashMap. For implementing queue using stacks:
Push op: TC & SC = O(1) as just appending
Pop and Peek op: TC = O(n) because of running a while loop to move elements to another stack and SC = O(n) because of using another stack
Am I right?

PR #1411, opened 2022-09-12
Completed design 2

gourishpisal08:


Hash_map.py
```diff
@@ -0,0 +1,18 @@
+#Time complexity is o(1)
+#Space complexity is o(1)
```
daredevilteja, line 2: SC is O(n)

Queue_Stack.py
```diff
@@ -0,0 +1,29 @@
+#Time complexity is O(n)
```
daredevilteja, line 1: TC is wrong

Queue_Stack.py
```diff
@@ -0,0 +1,29 @@
+#Time complexity is O(n)
+#space complexity is O(1)
```
daredevilteja, line 2: Sc is wrong

daredevilteja (changes requested):

daredevilteja:
Please do the requested changes

PR #1447, opened 2022-10-27
Completed Design Course 2, please review

katta332:


katta332:
Files are not correct so revoking PR

PR #1514, opened 2023-01-23
Completed Design-2: Queue implementation using stacks & HashMap Implementation - Please review

akhilreddy619:


osandeepreddy (commented):
Time complexity for both solutions is O(1)

osandeepreddy:
Time complexity for both solutions is O(1)

akhilreddy619:
For HashMap solution, I was confused how it could be O(1) since there is a loop checking for the key in the list. That is why I did not mention O(1) for hashmap. 

And for Queue implementation, I am pushing one stack data into another in all three methods. So, it cannot be O(1) right?

osandeepreddy:
Please write the optimal solution, where push operation is O(1) and pop is averagely O(1)

akhilreddy619:
Sandeep, I made a minor change to the Hashmap implementation. The size of the LinkedList array is increased to 10000. So in each index at most 100 elements will be stored in the linked list. So, TC will be O(100) -> O(1). This is taught in class. My implementation is the same as the one taught in the class except for the size of the map.

Also, made a minor change in queue implementation. Now push operation is O(1). But the pop and peek operations have the worst case as O(N) and on average it is O(1). The implementation is the same as taught in class.

PR #1694, opened 2023-08-09
implemented hashmap and queue using stack

fastakash:


Nandha2Tech (commented):
Good Job! In Problem2 empty() function, no need to pop from stack1 and push to stack2. It will result in O(n) time if stack1 is full. We can check whether both are empty and can return it.

Nandha2Tech:
Good Job! In Problem2 empty() function, no need to pop from stack1 and push to stack2. It will result in O(n) time if stack1 is full. We can check whether both are empty and can return it.

PR #1712, opened 2023-08-31
Completed Design-2 (Design HashMap and Implement Queue Using stack)

jailashah:
A few questions:
For Designing HashMap, I have solved using two solutions, 
1. Using Double Hashing - The test cases were failed while submitting on leetcode. 27 / 36 testcases passed.
Could you please suggest on what exactly is not correct in this code?
2. Using Array - For this the Space complexity should be O(1) because the array size would be constant of 1000001 or it should be O(N)?

### DIFF
--- a/DesignHashMapUsingArray.java
+++ b/DesignHashMapUsingArray.java
@@ -1,0 +1,53 @@
+//Implemented using Array technique
+// Time Complexity : O(1)
+// Space Complexity : O(1) //Question, whether this should O(1) because it is constant space of 1000001? or it should be O(N)
+// Did this code successfully run on Leetcode : YES
+// Any problem you faced while coding this : NO
+
+/***
+ * OUTPUT:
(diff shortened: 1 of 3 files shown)


Nandha2Tech (commented):
Great Work!
In DesignHashMapUsingDoubleHashing in line 61 to get buctet item use **key/bucketItems** instead of modulo to get the inner bucket. Making this change will pass all test cases.
Regarding SC for array of size greater size it will be O(N). (100000 is too big for a number to be considered as a constant)

Nandha2Tech:
Great Work!
In DesignHashMapUsingDoubleHashing in line 61 to get buctet item use **key/bucketItems** instead of modulo to get the inner bucket. Making this change will pass all test cases.
Regarding SC for array of greater size it will be O(N). (100000 is too big for a number to be considered as a constant)

PR #1715, opened 2023-08-31
Completed Design 2 - Please Review

apurvashinde357:


Nandha2Tech (commented):
Great Work!
For MyHashMap  TC would be O(1) since hashing is used to go the bucket and traverse of nodes in that bucket would be very less.
In MyQueue TC for push() and empty() would be O(1)

Nandha2Tech:
Great Work!
For MyHashMap  TC would be O(1) since hashing is used to go the bucket and traverse of nodes in that bucket would be very less.
In MyQueue TC for push() and empty() would be O(1)

PR #1729, opened 2023-09-18
Create hashmap.py

anaghayn:


Nandha2Tech (commented):
Great Work! Kindly include TC and SC. 
Please complete other problem too.

Nandha2Tech:
Great Work! Kindly include TC and SC. 
Please complete other problem too.

PR #1737, opened 2023-09-27
Create Implementing_Hashmap.py

KrishanuSinha:
Design a HashMap without using any built-in hash table libraries.

Implement the MyHashMap class:

MyHashMap() initializes the object with an empty map. void put(int key, int value) inserts a (key, value) pair into the HashMap. If the key already exists in the map, update the corresponding value. int get(int key) returns the value to which the specified key is mapped, or -1 if this map contains no mapping for the key. void remove(key) removes the key and its corresponding value if the map contains the mapping for the key.

### DIFF
--- a/Implementing_Hashmap.py
+++ b/Implementing_Hashmap.py
@@ -1,0 +1,41 @@
+class ListNode:
+    def __init__(self, key=-1,val=-1, next=None):
+        self.key = key
+        self.val = val
+        self.next = next
+        
+class MyHashMap:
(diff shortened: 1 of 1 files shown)


keshav-Gupta-IS (commented):
Good Work, Krishanu!! Keep it up. Also mention time and space complexities.

keshav-Gupta-IS:
Good Work, Krishanu!! Keep it up. Also mention time and space complexities.

PR #1747, opened 2023-09-28
Solved Design 2 Exercises

nithind-sudo:


ishika2741 (commented):
Good work! Kindly add TC and SC for second problem as well.

ishika2741:
Good work! Kindly add TC and SC for second problem as well.

PR #1751, opened 2023-09-29
Design-2 completed

praveennaini:


ImplementHashMap.java
```diff
@@ -0,0 +1,75 @@
+// Time Complexity : O(1) 
+// Space Complexity : O(1)
```
ishika2741, line 2: There would be some space that would be used here

ishika2741 (commented):
Well done!

ishika2741:
Well done!

PR #1764, opened 2023-10-12
Completed Design-2

shree-bd:
Please review


Nandha2Tech (commented):
Great Work! For design hash map TC would be O(1) even in worst case too.

Nandha2Tech:
Great Work! For design hash map TC would be O(1) even in worst case too.

PR #1777, opened 2023-11-10
Completed Design 2

BhagAneesh:


ishika2741 (commented):
Good work! Kindly add TC and Sc as well.

ishika2741:
Good work! Kindly add TC and Sc as well.

PR #1807, opened 2023-12-22
Completed design-2

keerthana-vijaykumar:


Problem-2.cpp
```diff
@@ -0,0 +1,93 @@
+// Time Complexity : O(N)
```
ishika2741, line 1: you are using a bucket size of 1000 which you can say as constant

ishika2741 (commented):
Well done!

ishika2741:
Well done!

Problem-2.cpp
keerthana-vijaykumar: okay. got it!

PR #1820, opened 2023-12-28
implemented hashmap using python lists

Humanvitha:


pvk3081 (commented):
One problem is missing

pvk3081:
One problem is missing

PR #1848, opened 2024-01-26
Queue using stack completed

kpranavreddy01:
Design 2 
Problem 1 completed.
I'm not sure about the space complexity. Learning that now. With whatever little knowledge I had have mentioned it there.


Nandha2Tech (commented):
Great Work! SC would be O(n). Kindly complete other problem also.

Nandha2Tech:
Great Work! SC would be O(n). Kindly complete other problem also.

PR #1985, opened 2024-07-27
Completed Design 2

poojts:


Hashmap.py
```diff
@@ -0,0 +1,55 @@
+#put, get and remove all are O(n) due to find function
```
ishika2741, line 1: but we consider average case complexity

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1988, opened 2024-07-28
completed the design 2

botlamanoj29:


Nandha2Tech (commented):
Great Work! In get of customHashMap, try to return -1 if the value does not exist.

Nandha2Tech:
Great Work! In get of customHashMap, try to return -1 if the value does not exist.

PR #1992, opened 2024-08-03
Update Sample.java

BoniPreetam:


Nandha2Tech (commented):
Great Work! Kindly include TC and SC and also update the other problem.

Nandha2Tech:
Great Work! Kindly include TC and SC and also update the other problem.

PR #2065, opened 2024-12-26
Design-2 : Completed.

Tanish-gupta1:


MyHashMap.java
```diff
@@ -0,0 +1,94 @@
+// Time Complexity :O(1)
+// Space Complexity : O(1)
```
Tanish-gupta1, line 2: This will be O(n) -> n would be no. of elements we're pushing.

Nandha2Tech (commented):
Great Work!

Nandha2Tech:
Great Work!

PR #2081, opened 2024-12-27
Design-2

Gauri-27:


Ques.py
```diff
@@ -0,0 +1,36 @@
+class MyQueue:
```
ishika2741, line 1: Please add time and space here.

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #2162, opened 2025-02-05
Completed Design-2

Akanksha1126:


Sample.java
```diff
@@ -1,7 +1,76 @@
-// Time Complexity :
-// Space Complexity :
+// Time Complexity : O(N)
```
ishika-gupta-sn, line 1: Check time complexity please . In case of design problems , always give complexity of individual functions

ishika-gupta-sn (commented):
Good work! Please submit other problem as well.

ishika2741:
Good work! Please submit other problem as well.

PR #2207, opened 2025-05-15
Design 2

arihantjain2198:
Problem 1 solved completely

Problem 2 only 26 test case solved. Unable to find issue in the approach

### DIFF
--- a/DesignHashMap.java
+++ b/DesignHashMap.java
@@ -1,0 +1,65 @@
+//problem 2: 
+// Time Complexity : Amortized O(1)
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : No
+// Any problem you faced while coding this : Only 26 test case passed but not sure why this logic is failing
+
+class MyHashMap {
+
+    private int[][] matrix;
+    int keys;
+    int keyItems;
+
+    private int hash1(int val) { // First hash function that do the modulo
+        return val % keys;
+    }
+
+    private int hash2(int val) {// First hash function that take do the division
+        return val / keyItems;
+    }
+
+    public MyHashMap() { // declare the variables
+        keys = 1000;
+        keyItems = 1000;
+        matrix = new int[keys][];
+
+    }
+
+    public void put(int key, int value) {
+        int valKey = hash1(key);//here use the hash functions
+        int valItem = hash2(key);
+
+        if (matrix[valKey] == null) { // if matrix is null no items yet for that key
+            matrix[valKey] = new int[keyItems]; // declare the array with keyItems
+        }
+
+        matrix[valKey][valItem] = value; // assign the value to the location based on hashing
+    }
+
+    public int get(int key) {
+
+        int valKey = hash1(key); //here use the hash functions
+        int valItem = hash2(key);
+
+        if (matrix[valKey] == null || matrix[valKey][valItem] == -1) {
+            return -1; // if no value for the key or removed return -1
+        }
+
+        return matrix[valKey][valItem];
+    }
+
+    public void remove(int key) {
+
+        int valKey = hash1(key);//here use the hash functions
+        int valItem = hash2(key);
+
+        if (matrix[valKey] == null) {
+            return; //if no key found return 
+        }
+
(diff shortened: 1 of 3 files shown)


DesignHashMap.java
```diff
@@ -0,0 +1,65 @@
+//problem 2: 
+// Time Complexity : Amortized O(1)
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : No
+// Any problem you faced while coding this : Only 26 test case passed but not sure why this logic is failing
+
+class MyHashMap {
+
+    private int[][] matrix;
```
sudhasr, line 9: Instead of int[][]  use   Integer[][]

DesignHashMap.java
```diff
@@ -0,0 +1,65 @@
...
+        return val % keys;
+    }
+
+    private int hash2(int val) {// First hash function that take do the division
+        return val / keyItems;
+    }
+
+    public MyHashMap() { // declare the variables
+        keys = 1000;
+        keyItems = 1000;
```
sudhasr, line 23: keyItems = 1001; // since the max key is 1000000, the hash2(1000000) will be 1000, so we need an array of size 1001 (0 to 1000 inclusive)

DesignHashMap.java
```diff
@@ -0,0 +1,65 @@
...
+        matrix = new int[keys][];
```
sudhasr, line 24: use Integer[keys][]

DesignHashMap.java
```diff
@@ -0,0 +1,65 @@
...
+
+    }
+
+    public void put(int key, int value) {
+        int valKey = hash1(key);//here use the hash functions
+        int valItem = hash2(key);
+
+        if (matrix[valKey] == null) { // if matrix is null no items yet for that key
+            matrix[valKey] = new int[keyItems]; // declare the array with keyItems
```
sudhasr, line 33: Integer[keyItems]

DesignHashMap.java
```diff
@@ -0,0 +1,65 @@
...
+
+        matrix[valKey][valItem] = value; // assign the value to the location based on hashing
+    }
+
+    public int get(int key) {
+
+        int valKey = hash1(key); //here use the hash functions
+        int valItem = hash2(key);
+
+        if (matrix[valKey] == null || matrix[valKey][valItem] == -1) {
```
sudhasr, line 44: if (matrix[valKey] == null || matrix[valKey][valItem] == null)

DesignHashMap.java
```diff
@@ -0,0 +1,65 @@
...
+
+        return matrix[valKey][valItem];
+    }
+
+    public void remove(int key) {
+
+        int valKey = hash1(key);//here use the hash functions
+        int valItem = hash2(key);
+
+        if (matrix[valKey] == null) {
```
sudhasr, line 56: if (matrix[valKey] != null) {
     matrix[valKey][valItem] = null;
  }

DesignHashMap.java
```diff
@@ -0,0 +1,65 @@
...
+            return; //if no key found return
```
sudhasr, line 57: no need to return here

DesignHashMap.java
```diff
@@ -0,0 +1,65 @@
...
+    public void remove(int key) {
+
+        int valKey = hash1(key);//here use the hash functions
+        int valItem = hash2(key);
+
+        if (matrix[valKey] == null) {
+            return; //if no key found return 
+        }
+
+        matrix[valKey][valItem] = -1; // if key found just add -1 to the value
```
sudhasr, line 60: not required as this is taken care in the if condition above

DesignHashMap.java
```diff
@@ -0,0 +1,65 @@
...
+
+        return;
```
sudhasr, line 62: not required

sudhasr (commented):
Good job on Queue using Stack. 
Good Try on the Design HashMap

sudhasr:
Good job on Queue using Stack. 
Good Try on the Design HashMap

PR #2290, opened 2025-06-20
Done Design 2

rk-peddibhotla:


super30admin:
Strengths:
- The implementation correctly follows the two-stack approach for queue operations.
- The student has included a main method for testing, which demonstrates good practice.

Areas for Improvement:
- While the comments are helpful, they could be more concise. For example, the explanation of the two-stack approach could be shortened.
- The main method could include more test cases to cover edge cases, such as calling pop or peek on an empty queue.

PR #2301, opened 2025-06-21
Done Design-2

Shreya1093:


super30admin:
- Strengths:

- Areas for Improvement:
  - In MyQueue, the `pop` method calls `peek` before popping, which is correct but could be noted in the comments for clarity.
  - In MyHashMap, the `find` method could be made private since it is a helper method and not meant to be exposed outside the class.
  - The comments in MyHashMap are slightly verbose and could be more concise.

PR #2303, opened 2025-06-21
Updated Design2-SAI

sindhoorpalla:


super30admin:
The student's implementation of MyQueue is excellent. It correctly uses two stacks to simulate a queue, with all operations functioning as expected. The code is clean, efficient, and follows best practices. One minor suggestion for improvement could be to add comments explaining the purpose of each method, especially the peek method's role in maintaining the queue's order, to enhance readability for others who might review the code.

PR #2306, opened 2025-06-22
completed implementqueueusingstack

JoshnaRavi04:


super30admin:
- The student's implementation of `MyQueue` is excellent. The logic is clear, and the operations are correctly implemented with the appropriate time and space complexities.
- One minor improvement could be to handle the case where `pop` is called on an empty queue. Currently, it returns `None`, but it might be better to return `-1` or raise an exception to match typical queue behavior.
- The comments in the code are helpful, but adding a brief docstring for the class and methods could further enhance readability and maintainability.

PR #2314, opened 2025-06-26
Added sol for Queue using Stack problem

satvikapadakanti:
Submitted solution for implementing Queue using Stack problem under Design-2

### DIFF
--- a/HashMap.java
+++ b/HashMap.java
@@ -1,0 +1,63 @@
+class MyHashMap {
+    private class Node {
+        int key, value;
+        Node next;
+
+        Node(int key, int value) {
+            this.key = key;
+            this.value = value;
+        }
+    }
+
+    private final int SIZE = 10000;
+    private Node[] buckets;
+
+    public MyHashMap() {
+        buckets = new Node[SIZE];
+    }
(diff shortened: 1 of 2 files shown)


super30admin:
Strengths:
- The implementation correctly uses two stacks to simulate a queue.
- The code is clean, with good variable naming and method structure.
- The peek method is efficiently reused in the pop method, reducing redundancy.

Areas for Improvement:
- While the code is correct, adding comments to explain the logic, especially the transfer between stacks in the peek method, would enhance readability for others.
- The empty method could be made more explicit by adding a comment that it checks both stacks to determine if the queue is empty.

super30admin:
- The student's solution is well-implemented and correctly solves the problem. The code is clean, efficient, and follows best practices.
- One minor improvement could be to add a check for empty queues in the pop() method before calling peek(), similar to the reference solution. However, this is not strictly necessary as peek() already handles the empty case by transferring elements if needed.
- The use of meaningful variable names (in and out) enhances readability.
- Overall, the student has demonstrated a good understanding of the problem and has provided a high-quality solution.

PR #2320, opened 2025-07-24
Completed Design-2: HashMap Implementation Using Separate Chaining

arpitha-nagaraj9791:


super30admin:
Strengths:
- The implementation is correct and follows the separate chaining approach.
- The code is modular and well-commented, making it easy to understand.
- The rehashing logic is correctly implemented to maintain efficiency.

Areas for Improvement:
- Consider lowering the rehashing threshold (e.g., 0.75) to reduce the likelihood of collisions before rehashing.
- Optimize the searchInLL method by using an iterator or enhanced for loop to improve performance.
- Add null checks or handle edge cases where keys might be null (though the problem constraints might not require this).

PR #2328, opened 2025-07-27
Done Design-2

ayushii08:


super30admin:
Strengths:
- The student has a good understanding of data structures and their implementations.
- The use of helper methods (`getPrev`) and dummy nodes shows advanced problem-solving skills.

Areas for Improvement:
- In `MyQueue`, the `peek()` method is called within `pop()` to ensure the front element is ready in `stack2`. While this works, it might be clearer to explicitly transfer elements in `pop()` if `stack2` is empty, rather than relying on `peek()`.
- The `MyHashMap` could benefit from a dynamic resizing mechanism to handle cases where the number of keys exceeds the initial bucket size, improving performance in such scenarios.

PR #2331, opened 2025-07-28
Nikhil Lohar_RN03JUL2025 : Complete the problem using Javascript

nikhil2297:


super30admin:
- The solution is well-implemented and follows the expected approach for designing a queue using stacks.
- The comments and explanations provided are clear and helpful for understanding the approach.
- One minor improvement could be to ensure consistency in the return value for pop when the queue is empty. The reference solution returns -1, which the student has followed, but it's worth noting that this might not be standard for all queue implementations (some might throw an exception).
- The code is clean and follows good practices, but the student could consider adding more detailed comments for the peek method to explain why it's called before pop.

PR #2334, opened 2025-07-28
Implement a HashMap

NehaKhan9753:


super30admin:
Strengths:
- The implementation correctly handles all required operations (put, get, remove) with appropriate time and space complexity.
- The use of a dummy node in each bucket simplifies edge case handling, which is a good practice.

Areas for Improvement:
- The `put` method could be slightly optimized by directly adding the new node when the bucket is empty, rather than creating a dummy node and then adding the new node. This would reduce one level of indirection in the best case.
- The space complexity comment could be clarified to distinguish between per-operation space complexity (O(1)) and overall data structure space complexity (O(n)).

PR #2335, opened 2025-07-29
Completed Design-2

sharmaarjun360:


super30admin:
- The implementation of MyQueue is correct and efficient. The use of two stacks to simulate a queue is well-executed.
- The code is clean and follows good practices. The variable names are clear, and the logic is easy to follow.
- A minor suggestion: Consider returning -1 instead of Int.MAX_VALUE when the queue is empty in pop and peek operations to maintain consistency with common practices and the reference solution.
- The comments in the code are helpful, but ensure they are up-to-date with any changes made during debugging (e.g., line 41 comment about checking both stacks).

PR #2343, opened 2025-08-20
Add MyQueue implementation using two lists

ravithakur2k:
Introduces a MyQueue class that simulates a queue using two lists. Includes methods for push, pop, peek, and empty, with example usage at the end.


super30admin:
Strengths:
- The solution correctly implements the required functionality.
- The code is readable and follows basic best practices.
- The comments provide a clear explanation of the time and space complexity.

Areas for Improvement:
- The pop and peek operations can be optimized to avoid transferring elements back to the original stack every time. The reference solution shows how to do this by only transferring elements when the output stack is empty.
- The variable names could be more descriptive to better reflect their purpose (e.g., `inSt` and `outSt` instead of `queue` and `stack`).
- The empty method could be simplified to return `not self.queue` directly, as it already returns a boolean.

PR #2345, opened 2025-08-20
Implement problem 1 - implement queue using stack and also HashMap

sarvesh-gaurishankar-sawant:


super30admin:
Strengths:
- The solution is correct and efficiently implements the queue using two stacks.
- Time and space complexity analysis is accurate and well-documented.
- Code is clean, well-commented, and follows good practices.

Areas for Improvement:
- The comment "SC: O(2 * n)" could be simplified to "SC: O(n)" since constants are typically dropped in Big-O notation.
- The empty() method could be simplified to a single line: `return input.isEmpty() && output.isEmpty();` instead of using an if-else statement.

sarvesh-gaurishankar-sawant:
Added implementation of HashMap, please review

PR #2348, opened 2025-08-20
Queue using Stack, HashMap Done

hp6318:


super30admin:
- Strengths: The student has correctly implemented the queue using two stacks, with appropriate time and space complexity. The code is well-commented and readable. The use of peek() in pop() is a good practice.
- Areas for Improvement: The empty() method can be simplified by directly checking if both stacks are empty, eliminating the need for a separate size variable. This would make the code cleaner and more efficient. Additionally, the sanity check in pop() and peek() could be removed if the problem guarantees valid operations, but this is a minor point.

ravithakur2k:
Added the implementation of HashMap

PR #2354, opened 2025-08-21
Design 2 HashMap

sathvikareddy18:


super30admin:
Strengths:
- The hash map implementation is clean and well-organized.
- The use of separate chaining with linked lists for collision handling is a good approach.
- The code is readable with appropriate comments and method names.

Areas for Improvement:
- The main issue is that the solution doesn't address the correct problem. Please carefully read the problem statement before implementing a solution.
- For the actual problem (implementing a queue using stacks), you would need to use two stacks (one for input and one for output) and carefully manage the transfer of elements between them to achieve queue operations.
- The ListNode class in your solution is unnecessary for the queue problem, but would be appropriate for the hash map problem you implemented.

PR #2359, opened 2025-08-22
implement queue using two stacks and HashMap solution java

Adikanu:


super30admin:
Strengths:
- The queue implementation correctly uses two stacks to achieve FIFO behavior.
- The HashMap implementation is well-organized with a helper method to manage the linked list operations.
- Comments and method names are clear and descriptive.

Areas for Improvement:
- The HashMap could benefit from a dynamic resizing mechanism to handle load factors better, though this is not strictly necessary for the problem.
- The helper method in the HashMap could be renamed to something more descriptive, like `findPreviousNode`, to better indicate its purpose.
- The queue's `peek` and `pop` methods could be optimized further by ensuring the `out` stack is only refilled when absolutely necessary, though the current implementation is already efficient.

PR #2370, opened 2025-08-29
Create hashMap

Harsha-Ramesh-Naik:


super30admin:
Strengths:
- The implementation correctly follows the hash map with chaining approach.
- Proper use of helper methods (_index) improves readability.
- The remove method handles edge cases well with a dummy node.

Areas for Improvement:
- The solution is for a different problem than what was asked (MyQueue vs MyHashMap).
- The hash map could benefit from dynamic resizing to maintain O(1) average time complexity as the number of elements grows.
- Adding docstrings to methods would improve documentation.
- The initial bucket size (1001) is arbitrary; it might be better to use a prime number or make it configurable.

PR #2384, opened 2025-10-03
Design2 submission

Bhumika-95:


super30admin:
Strengths:
- The implementation of MyQueue is correct and efficient.
- The use of a helper method (transferIfNeeded()) is a good practice to avoid code duplication.

Areas for Improvement:
- The file naming is inconsistent (HasMap.java vs Stack(Queue).java). It would be better to have consistent and clear file names.
- The comment about time complexity for pop() and peek() could be more explicit about the worst-case scenario (O(n)) when the out stack is empty and needs to be refilled.
- The empty() method could be documented to clarify that it checks both stacks.

PR #2388, opened 2025-10-05
HashMap and Queue Implementation

arjunsahas:


super30admin:
Your implementation of MyQueue is well-done and matches the reference solution in terms of correctness, time complexity, and space complexity. The code is clean and easy to understand. One small suggestion would be to add more detailed comments or JavaDoc for each method to make the code even more readable and maintainable. For example, explaining the peek() method's role in ensuring the outStack is populated before operations could be helpful for future readers.

PR #2391, opened 2025-10-06
Design 2

bharath6969:
Implemented the logic for queue implementation and also hash map using liner chaining 

1) Understood the concept of linear chaining using linked list but what's the difference  between double hashing and linked list traversing  from secondary datastructure one is contiguous we can get in o(1) because of the reference  and other non contiguous worst case O(n)

### DIFF
--- a/Hash_map.py
+++ b/Hash_map.py
@@ -1,0 +1,62 @@
+class MyHashMap:
+    class Node:
+        def __init__(self, key, val):
+            self.key = key
+            self.val = val
+            self.next = None
+
+    def __init__(self):
+        self.plength = 10000
+        self.ar = [None] * self.plength
+
+    def has1(self, key):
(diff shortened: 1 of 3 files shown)


super30admin:
Strengths:
- The student has implemented both problems correctly with optimal time and space complexity.
- The code is modular and readable, with good use of helper methods.
- The README.md provides a clear and concise explanation of the approach.

Areas for Improvement:
- In the MyQueue implementation, the empty() method can be simplified to `return not self.outstack and not self.instack` for better readability.
- In the MyHashMap implementation, the remove() method returns -1 in some cases, but the method signature suggests it should return None. This inconsistency should be fixed.
- Adding comments within the code to explain the logic, especially for the MyHashMap implementation, would further improve readability.

PR #2396, opened 2025-11-13
Create implementQueueUsingStack.py

shreyasja06:
Implement queue using stack using python


super30admin:
Strengths:
- The solution correctly implements the queue using two stacks, demonstrating a good understanding of the problem.
- The code is clean, readable, and follows good practices with descriptive variable names and comments.

Areas for Improvement:
- The docstrings could be more detailed to explain the purpose and return values of each method more clearly.
- The `peek` method could include a check for an empty queue, similar to the `pop` method, to handle edge cases more gracefully.
- The initial comment could be more concise and focused on the solution's approach rather than just the time complexities.

PR #2404, opened 2025-11-14
Completed Design-2

anushrihub:


super30admin:
Strengths:
- The implementation correctly uses two stacks to achieve FIFO behavior.
- The logic for transferring elements between stacks during pop and peek operations is correctly implemented.

Areas for Improvement:
- The peek method should return -1 (or raise an exception) when the queue is empty, rather than None, to maintain consistency with the pop method and typical queue behavior.
- The empty method could be used more consistently in the peek method to check for an empty queue before proceeding.

PR #2410, opened 2025-11-18
Add files via upload

Srijha-Kalyan:


super30admin:
- Problem1.py: Excellent job! Your implementation is clean, efficient, and handles edge cases well. The use of a helper function (`_shift_stacks`) is a good practice.
- Problem2.py: Your approach is conceptually sound, but there's a critical bug in the initialization of the hash table. Instead of `[ListNode] * self.size`, you should initialize each bucket with a new `ListNode` (e.g., `[ListNode() for _ in range(self.size)]`). This ensures each bucket is independent. Also, consider adding a check in `remove` to handle the case where the key is not found.

PR #2411, opened 2025-11-24
Done Design-2

gurneetk186:


super30admin:
Strengths:
- Both solutions are well-commented, making the logic easy to follow.
- The use of helper methods (moveInToOut and find) improves modularity and readability.
- The hash map implementation handles edge cases (like empty buckets) gracefully.

Areas for Improvement:
- In Problem1.java, the moveInToOut method could be called peek in the pop method to ensure consistency with the reference solution, though this is a minor stylistic point.
- In Problem2.java, the SIZE constant could be documented to explain why 10000 was chosen, though this is not critical for correctness.

PR #2416, opened 2025-12-25
Completed hashmap with double hashing and implemented queue

kkhevna9252:


super30admin:
- The student's MyQueue implementation is correct and efficient, matching the reference solution in both functionality and complexity analysis.
- The inclusion of MyHashMap in the submission is confusing since it's not part of the problem being evaluated (Design-2). Focus on submitting only the relevant code for the problem at hand.
- The Kotlin code is clean and follows good practices. The use of Java's Stack class in Kotlin is acceptable, though Kotlin's native collections could also be considered.
- The comments about time complexity are accurate and well-placed.

PR #2420, opened 2025-12-25
Completed Design 2

thondaha:


super30admin:
Strengths:
- Good use of comments explaining the approach and complexity
- Proper separation of concerns between classes
- Correct implementation of the queue using two stacks
- Appropriate use of data structures (linked list for hash map buckets)

Areas for Improvement:
- The remove method in MyHashMap has a logical error. It should check cur.next.key == key instead of cur.key == key in the while loop.
- Consider adding null checks or input validation for edge cases
- The hash map could benefit from implementing resizing when the load factor becomes too high
- The test file could be more comprehensive, testing edge cases like empty queues/maps
- In MyHashMap, the while loop condition in remove is incorrect (while(cur.next != null) should be while(cur != null))

PR #2421, opened 2025-12-26
Done with Design-2

Mukul585:


super30admin:
Strengths:
- The implementation correctly uses two stacks to simulate a queue, adhering to the FIFO principle.
- The time and space complexities are optimal for this problem.

Areas for Improvement:
- The logic for transferring elements from inStack to outStack is duplicated in both pop and peek methods. You could call peek within pop to avoid this duplication, making the code slightly more maintainable.
- While the comments are helpful, they could be more detailed in explaining why the transfer from inStack to outStack is necessary for achieving FIFO behavior.

PR #2422, opened 2025-12-26
Implemented Design-2

sneha-tambade:


super30admin:
Strengths:
- The student has correctly implemented the queue using two stacks, demonstrating a good understanding of the problem.

Areas for Improvement:
- The `Console.WriteLine(outStack.Peek());` in the `Peek` method seems unnecessary for the functionality and could be removed to avoid side effects.
- While the comments are helpful, they could be more detailed to explain why the two-stack approach works (FIFO using LIFO structures).

PR #2428, opened 2026-01-07
Design 2 solutions

dbkuppagiri:


super30admin:
Strengths:
- Good use of two stacks to implement the queue functionality.
- Clear and readable code with appropriate comments.
- Correct implementation of the `push` and `empty` methods.

Areas for Improvement:
- The `peek` method should handle the case when both stacks are empty (e.g., return -1 or throw an error).
- The `pop` method should check if the queue is empty before attempting to pop (similar to the reference solution).
- Consider adding more detailed comments about the time complexity of each method.

PR #2433, opened 2026-01-13
Design-2: Queue using stack HW

Amaan-29:


super30admin:
Strengths:
- The implementation correctly uses two stacks to simulate a queue, adhering to the problem constraints.
- The code is clean, with good variable naming and logical structure.
- The time and space complexity are optimal for this problem.

Areas for Improvement:
- While the problem guarantees valid calls to pop and peek, adding a check for empty queues in these methods could make the code more robust for general use.
- Consider adding comments to explain the logic, especially for someone who might be new to this concept.

PR #2436, opened 2026-02-04
Added MyQueue and MyHashMap solutions

PritPanchani:


super30admin:
Strengths:
- The solution for `MyQueue` is correct and efficient.
- The comments and approach description are helpful.

Areas for Improvement:
- The submission includes an unrelated solution for `MyHashMap`. In the future, ensure that only the required solution is submitted for the specific problem. This helps in avoiding confusion and keeps the focus on the problem at hand.
- Although not necessary, consider adding null checks or handling edge cases explicitly, but the current implementation already handles them correctly.

super30admin:
Your MyQueue solution is well-implemented and efficient. It correctly uses two stacks to simulate a queue with amortized constant time operations. The code is clean and easy to understand. However, note that you included a solution for MyHashMap in the same submission, which is not part of the problem. Make sure to only submit the code relevant to the problem. For the MyQueue solution, you might consider adding a comment to explain the approach briefly, but it's not necessary as the code is self-explanatory.

PR #2440, opened 2026-02-05
Add Python implementations for LeetCode 232 and 706

rishigoswamy:


super30admin:
For leetcode232.py (Queue using Stacks):
- The solution is correct and follows the standard approach.
- The code is clean and well-commented.
- One minor improvement: in the `pop` method, you can call `peek` to ensure the outputStack has elements, which might make the code slightly more concise (as in the reference solution). However, your current implementation is also correct and clear.

For leetcode706.py (Hash Map):
- The solution is correct and uses separate chaining with linked lists.
- In the `put` method, the variable `presentInLinkedList` is unnecessary. You can traverse the linked list and update the value if the key exists. If you reach the end without finding the key, you add a new node. This avoids using a flag variable.
- Example improvement for `put`:
  ```
  def put(self, key: int, value: int) -> None:
      idx = self.hashFunction(key)
      curr = self.array[idx]
      while curr.next:
          if curr.next.key == key:
              curr.next.value = value
              return
          curr = curr.next
      curr.next = Node(key, value)
  ```
- However, note that your current implementation uses a dummy node at each bucket, which is good. The above suggestion is just to simplify the code.
- Also, in the `remove` method, you are starting from the dummy node and checking `nxt.key`, which is correct. But ensure that you are not missing the case where the key might be in the dummy node? Actually, the dummy node has key None, so it won't match. So it's safe.

Overall, both solutions are excellent. The code is well-organized and easy to understand.

super30admin:
Your solution for the "Create Queue using Stacks" problem is well-implemented. The code is clean and follows the standard approach. The comments and docstring are helpful. However, note that in the `pop` method, if the queue is empty, your code would crash because you don't handle the case when both stacks are empty. Although the problem states that all calls to `pop` and `peek` are valid (so they won't be called on an empty queue), it is still good practice to include a safety check, such as returning -1 or raising an exception with a message. But since the problem constraints guarantee valid calls, your solution is acceptable.

One minor improvement: you can avoid duplicating the transfer logic in both `pop` and `peek` by having `peek` handle the transfer and then `pop` call `peek` to ensure the output stack is populated, as done in the reference solution. This reduces code duplication.

For example, the reference solution does:
```python
def pop(self) -> int:
    if self.empty():
        return -1
    self.peek()
    return self.outSt.pop()
```

This way, the transfer logic is only written in `peek`. You might consider this for cleaner code.

Overall, great job!

PR #2444, opened 2026-02-06
Add HashMap and QueueStack implementations

Balaviknesh:


super30admin:
Strengths:
- You have correctly understood the approach of using two stacks to implement a queue.
- The transfer of elements from inStack to outStack when outStack is empty is correctly implemented.
- The time and space complexity are optimal.

Areas for improvement:
- The class should be named "MyQueue" as per the problem statement. The example code in the problem uses "MyQueue", and the test cases will expect that class name.
- You should handle the case when the queue is empty in pop and peek. Although the problem states that all calls are valid, it is good practice to handle edge cases to avoid runtime exceptions. You can either throw an exception with a message or return a sentinel value (like -1) as in the reference solution. However, since the problem says all calls are valid, you can also assume it's safe, but without handling, it will crash.
- There is code duplication in pop and peek for transferring elements. You can refactor by having a helper function or by having peek handle the transfer and then pop uses peek. This makes the code cleaner and avoids duplication.
- The solution for HashMap is included in the same file, which is not required for this problem. Please ensure you only submit the relevant code for the problem.

Suggested improvements for the QueueStack class (renamed to MyQueue):
```
class MyQueue {
    private val inStack = mutableListOf<Int>()
    private val outStack = mutableListOf<Int>()

    fun push(x: Int) {
        inStack.add(x)
    }

    fun pop(): Int {
        peek() // Ensure outStack has the front element
        return outStack.removeAt(outStack.size - 1)
    }

    fun peek(): Int {
        if (outStack.isEmpty()) {
            while (inStack.isNotEmpty()) {
                outStack.add(inStack.removeAt(inStack.size - 1))
            }
        }
        return outStack[outStack.size - 1]
    }

    fun empty(): Boolean {
        return inStack.isEmpty() && outStack.isEmpty()
    }
}
```
This reduces duplication and ensures that both pop and peek behave consistently.

super30admin:
Strengths:
- The solution correctly implements the queue using two stacks with the required amortized time complexity.
- The code is straightforward and easy to understand.

Areas for improvement:
1. Class Naming: The problem requires the class to be named "MyQueue". Rename "QueueStack" to "MyQueue" to match the problem specification.
2. Code Duplication: The logic for transferring elements from inStack to outStack is duplicated in both pop and peek. Consider refactoring by having pop call peek (like the reference solution) or by creating a helper function to transfer elements. This would make the code more maintainable.
3. Kotlin Idioms: You can use Kotlin's built-in functions to make the code more concise. For example, instead of using mutableListOf, you could use ArrayDeque which is more efficient for stack operations (though mutableListOf is acceptable). Also, consider using `last()` and `removeLast()` for better readability when accessing the last element.
4. Edge Case Handling: The pop function does not handle the case when the queue is empty. The problem states that all calls to pop and peek are valid, so it might not be necessary. But for robustness, you might consider throwing an exception or returning a sentinel value if empty (though the problem constraints say all calls are valid).

Suggested Refactoring for pop and peek:
```
fun pop(): Int {
    peek() // This ensures outStack has the front element if exists
    return outStack.removeAt(outStack.size - 1)
}

fun peek(): Int {
    if (outStack.isEmpty()) {
        while (inStack.isNotEmpty()) {
            outStack.add(inStack.removeAt(inStack.size - 1))
        }
    }
    return outStack[outStack.size - 1]
}
```
This avoids duplication and is cleaner.

PR #2453, opened 2026-04-08
Implemented Queue using Stack

LeelaTotapally:


super30admin:
###  Create Queue using Stacks (QueueUsingStack.java)
Your solution is well-structured and follows the standard approach for implementing a queue using two stacks. The code is readable and includes comments explaining the approach, which is good.

Strengths:
- Correct implementation of the queue operations.
- Good use of two stacks to achieve FIFO behavior.
- Proper handling of the transfer of elements from `inStack` to `outStack` only when needed.

Areas for improvement:
1. The `reverse` method checks if `outStack` is empty again, which is unnecessary because the method is called only when `outStack` is empty. You can simplify the method by removing the condition and directly transferring elements from `inStack` to `outStack` until `inStack` is empty.
2. Consider renaming the `reverse` method to something more descriptive, like `transferIfNeeded`, to better reflect its purpose. However, this is optional and does not affect correctness.
3. In the `pop` method, after calling `reverse`, you directly return `outStack.pop()`. This is correct, but note that if the queue is empty, `outStack.pop()` might throw an exception. However, the problem states that all calls to `pop` and `peek` are valid, so it's acceptable. But for robustness, you might want to handle the empty case, but it's not required here.

Here is a slightly optimized version of your code:
```java
class MyQueue {
    Stack<Integer> inStack = new Stack<>();
    Stack<Integer> outStack = new Stack<>();

    public MyQueue() {}
    
    public void push(int x) {
        inStack.push(x);
    }

    private void transferIfNeeded() {
        if (outStack.isEmpty()) {
            while (!inStack.isEmpty()) {
                outStack.push(inStack.pop());
            }
        }
    }
    
    public int pop() {
        transferIfNeeded();
        return outStack.pop();
    }
    
    public int peek() {
        transferIfNeeded();
        return outStack.peek();
    }
    
    public boolean empty() {
        return inStack.isEmpty() && outStack.isEmpty();
    }
}
```

Overall, your solution is correct and efficient. The minor redundancy in the `reverse` method does not affect the correctness or performance significantly.

VERDICT: PASS

---

### Implement Hash Map
It seems there was a mix-up in the solution you submitted. The code you provided is for implementing a queue using two stacks (LeetCode problem 232), but the problem you were asked to solve is about implementing a hash map (LeetCode problem 706). 

To correctly solve the hash map problem, you should design a class `MyHashMap` that uses an array of linked lists (or another method) to handle key-value pairs. The reference solution provided uses a common approach with chaining to resolve collisions. Here are some key points to consider for your solution:

1. **Initialization**: Initialize an array of nodes (or a similar structure) with a fixed number of buckets (e.g., 1000) to store the key-value pairs.
2. **Hash Function**: Use a simple hash function (e.g., `key % buckets`) to determine the index in the array for a given key.
3. **Handling Collisions**: Use linked lists in each bucket to handle collisions. Each node in the linked list should store the key, value, and a reference to the next node.
4. **Put Method**: 
   - Calculate the index for the key.
   - If the bucket is empty, create a dummy head node and then add the new node.
   - Otherwise, traverse the linked list to check if the key already exists. If it does, update the value; if not, append a new node.
5. **Get Method**:
   - Calculate the index for the key.
   - Traverse the linked list at that index to find the key. Return the value if found, else return -1.
6. **Remove Method**:
   - Calculate the index for the key.
   - Traverse the linked list to find the node with the key and remove it by adjusting the pointers.

Your current solution does not address any of these requirements. Please review the problem statement again and implement the `MyHashMap` class as described.

VERDICT: NEEDS_IMPROVEMENT

PR #2458, opened 2026-04-09
Done Design-2

PavanKumarBollam:


super30admin:
###  Create Queue using Stacks (Sample.java)
Your solution is well-implemented and follows the standard approach for this problem. Here are some points to consider:

Strengths:
- You have correctly used two stacks to simulate the queue operations.
- The amortized time complexity is O(1), which is efficient.
- The code is clean, readable, and well-commented.
- You have included a private helper method to avoid code duplication, which is a good practice.

Areas for Improvement:
- Since the problem states that all calls to `pop` and `peek` are valid, you don't need to check for an empty queue in these methods. However, it's not wrong to have this check for safety. Alternatively, you could throw an exception if the queue is empty, but given the constraints, it's acceptable to return -1 as a fallback.
- The variable names `mainStack` and `outStack` are clear, but in some implementations, these are named `inputStack` and `outputStack` for even clearer intent. However, your names are fine.

Overall, your solution is correct and efficient. Keep up the good work!

VERDICT: PASS

---

### Implement Hash Map
It appears you have submitted a solution for the wrong problem. The problem was to design a HashMap, but you implemented a queue using two stacks. This is a critical error. Please review the problem statement carefully. 

For the HashMap problem, you need to store key-value pairs and support put, get, and remove operations. A common approach is to use an array of linked lists (chaining) to handle collisions. Each bucket in the array corresponds to a hash index, and each node in the linked list stores a key-value pair. 

Here are some steps to correct your solution:
1. Define a class for nodes that hold key, value, and a next pointer.
2. Initialize an array of nodes (buckets) with a certain size (e.g., 1000).
3. Implement a hash function to map keys to an index in the array.
4. For put: 
   - Calculate the hash index.
   - If the bucket is empty, create a dummy head node and add the new node.
   - Otherwise, traverse the linked list to check if the key exists. If it does, update the value; if not, add a new node.
5. For get:
   - Calculate the hash index.
   - Traverse the linked list to find the key. Return the value if found, else -1.
6. For remove:
   - Calculate the hash index.
   - Traverse the linked list to find the key and remove the node.

You should also consider handling resizing if the number of elements grows too large, but given the constraints (at most 10^4 calls), it might not be necessary.

Please reattempt the problem with the correct requirements in mind.

VERDICT: NEEDS_IMPROVEMENT

PavanKumarBollam:
In the README.md file, there were two problems: Implement Queue using Stacks and Design HashMap. Previously, I had already implemented the Queue using Stacks, but I had not implemented the HashMap.

Now, I have completed the HashMap implementation and also updated the Queue using Stacks section with the correct file name

PR #2460, opened 2026-04-11
Implement Queue with Stacks

hiteshmadapathi:


super30admin:
###  Create Queue using Stacks (ImplementQueueWithStacks.py)
Strengths:
- The solution correctly implements the queue operations using two stacks.
- The amortized time complexity is achieved by lazy transfer from st1 to st2.
- The code is mostly readable and well-commented.

Areas for improvement:
1. The pop method should ideally check if the queue is empty before popping, but since the problem states all calls are valid, it's acceptable. However, for robustness, consider adding a check (like the reference solution does) to return -1 or raise an exception if empty.
2. The peek method transfers elements every time st2 is empty, which is correct, but note that if peek is called multiple times without st2 being empty, it doesn't need to transfer. Your implementation does this correctly, but the reference solution avoids redundant transfers by having a separate peek method that only transfers when needed. Your implementation is actually similar to the reference.
3. The empty method can be simplified to `return not self.st1 and not self.st2` (like the reference) instead of using an if-return.
4. Variable names: st1 and st2 are acceptable, but more descriptive names like inStack and outStack (as in the reference) could improve readability.
5. In the pop method, after transferring, you pop from st2. This is correct. However, note that the reference solution calls peek() first to ensure st2 has the front element, then pops. Your implementation does the same without the explicit call to peek, which is efficient.

Overall, the solution is correct and efficient. With minor improvements in code style, it would be excellent.

VERDICT: PASS

---

### Implement Hash Map
It seems there might have been a mix-up in the problem you were solving. The problem was to implement a HashMap, but you provided a solution for implementing a queue using stacks. Please review the problem statement again: you need to design a HashMap class with methods `put`, `get`, and `remove`.

For the HashMap implementation, you should consider using an array of linked lists (or another method) to handle collisions. The reference solution provided uses a hash function to index into an array and handles collisions with linked lists. Each bucket in the array points to a linked list of nodes that store key-value pairs.

Key points for the HashMap:
- Initialize an array with a fixed number of buckets.
- Implement a hash function (e.g., key % number_of_buckets) to determine the index.
- For `put`, find the bucket and then traverse the linked list to update an existing key or add a new node.
- For `get`, find the bucket and traverse the linked list to find the key.
- For `remove`, find the bucket and remove the node with the given key.

Please try to implement the correct solution. If you have questions about the HashMap implementation, feel free to ask.

VERDICT: NEEDS_IMPROVEMENT

PR #2462, opened 2026-04-12
design-2

spencerkrebs:


super30admin:
###  Create Queue using Stacks (queue-using-stacks.py)
Strengths:
- The solution correctly uses two stacks to simulate a queue, with `inStack` for push operations and `outStack` for pop and peek operations.
- The time complexity for push is O(1), and for pop and peek, it is amortized O(1) because each element is moved from `inStack` to `outStack` at most once.
- The space complexity is O(n), which is optimal for this problem.
- The code is clean and well-commented, making it easy to understand.

Areas for Improvement:
- The `empty` method can be simplified. Instead of checking the lengths of both stacks and returning a boolean based on that, you can directly return the result of the condition `not self.inStack and not self.outStack`. This is more concise and efficient.
- While the comments are helpful, they are not strictly necessary since the code is self-explanatory. However, they are a good practice for clarity.

Suggested Improvement for `empty`:
```python
def empty(self) -> bool:
    return not self.inStack and not self.outStack
```

This change makes the method more efficient and aligns with the reference solution.

Overall, the solution is excellent and only requires a minor improvement in the `empty` method for conciseness.

VERDICT: PASS

---

### Implement Hash Map (design-hashmap.py)
Your solution has a good start with the use of chaining and dummy nodes for each bucket. However, there are some issues that need to be addressed:

- In the `get` method, you are starting from the dummy node (which has key=-1) and checking the current node's key. Since the dummy node is always present and its key is -1, you will never match the target key. Instead, you should traverse the list by checking the next node's key, similar to how you do in the `put` and `remove` methods. For example, in `get`, you should do:
  ```python
  cur = self.map[self.hash(key)]
  while cur.next:
      if cur.next.key == key:
          return cur.next.val
      cur = cur.next
  return -1
  ```

- Similarly, in the `remove` method, you are correctly checking the next node's key, but you are not handling the case where the key might be the first node after the dummy. Your current `remove` method should work because you are checking `cur.next.key`, but note that you are starting from the dummy node. However, you have an extra condition `while cur and cur.next` which is correct.

- In the `put` method, you are correctly traversing the list to update an existing key or append a new node. However, you should note that the dummy node is always present, so you don't need to check for the head being null.

- You initialized the map with 1000 dummy nodes. This is acceptable, but you should consider that the problem constraints allow up to 10^4 operations, so 1000 buckets might be too low. The reference solution uses 1000 buckets as well, so it's acceptable.

- Your hash function uses `key % len(self.map)`, which is correct. However, note that `len(self.map)` is 1000, so the modulo operation will give a value between 0 and 999.

- It's good practice to use a dummy node to simplify the linked list operations, but you need to ensure all methods are consistent in their traversal.

To fix your solution:
1. Correct the `get` method to traverse by checking the next node.
2. Ensure the `remove` method correctly removes the node by adjusting pointers.
3. Test your solution with the provided example to ensure it works.

Here is a corrected version of your code:
```python
class Node:
    def __init__(self, key=-1, val=-1, next=None):
        self.key = key
        self.val = val
        self.next = next

class MyHashMap:
    def __init__(self):
        self.map = [Node() for _ in range(1000)]
    
    def hash(self, key):
        return key % len(self.map)
    
    def put(self, key, value):
        index = self.hash(key)
        cur = self.map[index]
        while cur.next:
            if cur.next.key == key:
                cur.next.val = value
                return
            cur = cur.next
        cur.next = Node(key, value)
    
    def get(self, key):
        index = self.hash(key)
        cur = self.map[index]
        while cur.next:
            if cur.next.key == key:
                return cur.next.val
            cur = cur.next
        return -1
    
    def remove(self, key):
        index = self.hash(key)
        cur = self.map[index]
        while cur.next:
            if cur.next.key == key:
                cur.next = cur.next.next
                return
            cur = cur.next
```

VERDICT: NEEDS_IMPROVEMENT

PR #2463, opened 2026-04-14
Completed Design 2 Problem

pratikb0501:


super30admin:
###  Create Queue using Stacks (Problem_1.py)
Your solution is well-structured and correctly implements the queue using two stacks. Here are some points to consider:
1. **Correctness**: Your solution correctly handles the FIFO operations. The `push`, `pop`, `peek`, and `empty` methods are implemented as required.
2. **Time Complexity**: Your solution has the same amortized O(1) time complexity as the reference solution. This is efficient and meets the problem requirements.
3. **Space Complexity**: Your solution uses O(n) space, which is optimal for this problem.
4. **Code Quality**: The code is readable and follows good practices. However, you should avoid including test code within the class file. It's better to separate test cases from the implementation, especially when submitting code for evaluation.
5. **Edge Cases**: Although the problem states that all calls to `pop` and `peek` are valid, it's generally good practice to handle edge cases (like an empty queue) to make the code more robust. However, for this problem, it is not strictly necessary.

Overall, your solution is excellent. Just remember to remove any test code when submitting for evaluation.

VERDICT: PASS

---

### Implement Hash Map (Problem_2.py)
Your solution has a good idea of using a two-level array to avoid collisions, which ensures O(1) time complexity for operations. However, there is a critical issue with the _hash2 function and the size of the subbucket arrays.

1. The _hash2 function should be key // self.subbuckets, but this can generate indices from 0 to 1000 (for keys up to 10^6). For bucket0, you allocated an array of size 1001 (indices 0 to 1000), which is correct. But for other buckets, you allocated arrays of size 1000 (indices 0 to 999), which cannot handle keys that have _hash2 value 1000. For example, key=1000000: 
   - _hash1(1000000)=0, which uses the array of size 1001 -> index 1000 is valid.
   - key=1000001: _hash1(1000001)=1, and _hash2(1000001)=1000, but the array for bucket1 has size 1000 -> index 1000 is out of bounds.

2. To fix this, you should allocate the same size for all buckets. Since the maximum value of _hash2(key) is 1000 (for key=10^6), all subbucket arrays should have length 1001. You can change the initialization to:
   if currentBucket == 0:
        self.my_map[currentBucket] = [-1] * (self.subbuckets + 1)
   else:
        self.my_map[currentBucket] = [-1] * (self.subbuckets + 1)   # same for all

   But actually, you can simply always allocate arrays of length 1001. There is no need to treat bucket0 differently.

3. Alternatively, you can change the _hash2 function to key % self.subbuckets to generate indices in the range [0, 999]. But then you would have collisions within a bucket. However, that would require a different approach (like chaining or open addressing) within the bucket. Your current approach is actually a direct addressing scheme with two levels, which requires that the _hash2 function uniquely identifies the key within the bucket. But note: _hash1(key)=key % 1000, and _hash2(key)=key//1000. This pair (hash1, hash2) is unique for each key in the range [0, 10^6] because key = hash1 + 1000 * hash2. So it is a perfect hash for keys up to 10^6. Therefore, you should ensure that the subbucket array has enough slots to cover all possible hash2 values (0 to 1000). So the subbucket array should have 1001 slots for every bucket.

4. Also, note that the problem constraints say keys are in [0, 10^6], so the maximum key is 1000000. For key=1000000, hash2=1000. So you need to allocate 1001 slots for every bucket.

5. Another improvement: instead of initializing all subbucket arrays upfront, you could initialize them on demand. But your current approach initializes the subbucket array only when the bucket is first accessed. This saves memory if many buckets are unused.

6. In the remove method, you set the value to -1. This is correct because the problem says values are non-negative.

7. In the get method, you check if the value is not -1. But if someone stored value=0, it would be returned correctly because 0 != -1. So that is correct.

8. However, the problem requires that the map should be able to handle any number of keys up to 10^4 calls. But your solution preallocates 1000 buckets and each bucket has 1001 slots, which is 1000*1001=1,001,000 slots. This is acceptable in terms of memory (about 8MB if integers are 8 bytes) but might be considered wasteful if only a few keys are stored.

9. Overall, with the fix to allocate 1001 slots for every bucket, your solution would be correct. Without this fix, it will crash for keys >=1000000.

VERDICT: NEEDS_IMPROVEMENT

PR #2465, opened 2026-04-28
Design-2 Homework Solution

richapatil:


super30admin:
###  Create Queue using Stacks (232. Implement Queue using Stacks/MyQueue.java)
Strengths:
- Correct implementation of the two-stack queue.
- Good variable names.
- Includes complexity analysis.

Areas for improvement:
- The check for empty in pop() is not strictly necessary since the problem guarantees valid calls. But it's not wrong.
- In the comment, "TC = O(n)" might be misinterpreted. It should be amortized O(1) per operation. Overall, for n operations, it's O(n), but per operation it's constant on average.

VERDICT: PASS

---

### Implement Hash Map (706. Design HashMap/MyHashMap.java)
The student's solution attempts to implement a hash map using a two-dimensional array approach. However, there are several issues that prevent it from being correct and efficient.

1. **Correctness**: The solution has a critical flaw in the handling of keys. The primary hash function uses `key / primaryBucket`, which is integer division. Since `primaryBucket` is 1000, this means that for keys from 0 to 999, `primaryIndex` will be 0; for keys 1000 to 1999, `primaryIndex` will be 1; and so on. However, the secondary hash function uses `key % secondaryBucket`, which is modulo 1000. This leads to collisions for keys that are not in the same primary bucket but have the same modulo 1000. For example, keys 1 and 1001 both have `secondaryIndex = 1`, but they are in different primary buckets. However, the problem requires storing key-value pairs without such collisions. Additionally, the solution does not handle the case where the same key is inserted multiple times (it should update the value), but because of the way the array is allocated, it might overwrite values incorrectly. Also, the removal sets the value to -1, but this does not actually remove the key; it just sets the value to -1, which might be confused with a valid value if the problem allowed negative values (which it doesn't, but still, the problem requires removal of the mapping). However, the problem constraints say that values are non-negative, so setting to -1 is acceptable for "not found", but the issue is that the key is still present in the array (with value -1) and the array is allocated for the entire bucket, which is inefficient.

2. **Time Complexity**: The operations put, get, and remove are O(1) in the best case, which is good. However, due to the incorrect handling of collisions, the solution might not work correctly for all inputs.

3. **Space Complexity**: The solution allocates arrays of size 1000 or 1001 for each primary bucket. However, it only allocates a primary bucket when a key in that bucket is inserted. The total number of primary buckets is 1001 (from index 0 to 1000). The worst-case space is about 1001 * 1000 = 1,001,000 integers, which is acceptable given the constraints (keys up to 10^6). But note that the problem says "at most 10^4 calls", so the number of keys is limited, but the solution allocates arrays for entire buckets even if only one key in that bucket is used. This is inefficient in terms of space compared to the reference solution which uses linked lists and only allocates nodes for actual keys.

4. **Code Quality**: The code is readable and well-structured. The methods are clearly defined. However, the comments about time and space complexity are incorrect: the space complexity is O(10^6) which is constant, but it is actually O(1) because the array sizes are fixed. However, the student claims "SC - O(10^6)" which is misleading because it is constant space, but also the constant is large.

5. **Efficiency**: The solution is inefficient in space because it pre-allocates arrays for secondary buckets even if they are not fully utilized. Also, the collision handling is flawed, so it might not work for all keys.

The main issue is that the hash functions do not uniquely identify keys. For example, consider key=0: primaryIndex = 0/1000=0, secondaryIndex=0%1000=0. Key=1000: primaryIndex=1000/1000=1, secondaryIndex=1000%1000=0. So they are stored in different places, which is good. But key=1 and key=1001: primaryIndex for 1 is 0, for 1001 is 1; secondaryIndex for both is 1. So they are stored in different primary arrays but at the same secondary index. This is correct because they are different keys. However, the problem is that the secondary array is allocated for the entire bucket, and the secondary index is computed by modulo, so for a given primary bucket, the secondary index is unique. So actually, the solution might work correctly for distinct keys? But wait: what if two keys have the same primaryIndex and the same secondaryIndex? For example, key=0 and key=1000? They have the same secondaryIndex (0) but different primaryIndex. So no collision. But within the same primary bucket, the secondaryIndex is unique because it is computed by modulo. So for keys in the same primary bucket, the secondaryIndex is unique. Therefore, the solution should work correctly for distinct keys. However, the problem requires that when the same key is put again, the value should be updated. This is handled because the same key will always have the same primaryIndex and secondaryIndex, so it will be stored in the same location. So the put method will overwrite the value. The get method returns the value at that location, and remove sets it to -1. So it should work correctly for the given constraints (keys and values are non-negative, and keys are integers). However, there is one more issue: the primaryIndex for key=1000000? The primaryBucket is 1000, so primaryIndex = 1000000 / 1000 = 1000. The storage array has size 1001 (indices 0 to 1000), so it is valid. The secondaryIndex = 1000000 % 1000 = 0. So it is stored at storage[1000][0]. But note that for primaryIndex=0, the student allocates an array of size 1001 (because of the condition if primaryIndex==0), and for other indices, size 1000. So for key=1000000, which has primaryIndex=1000, the array allocated is of size 1000. So secondaryIndex=0 is within [0,999]. So it is valid. However, what about key=0? It is stored at storage[0][0] and the array for primaryIndex=0 has size 1001, so index 0 is valid. So the solution might actually work correctly for all keys in the range [0, 10^6]. But wait: the problem says keys can be in [0,10^6]. So the maximum key is 10^6. The primaryIndex for key=10^6 is 1000. The secondaryIndex is 0. So it is stored correctly. However, the solution does not handle keys that are not in [0,10^6]? The problem says keys are in that range, so it is acceptable.

But there is a critical issue: the secondary hash function for keys in the same primary bucket: for primaryIndex=0, the keys are from 0 to 999. The secondaryIndex is key % 1000, which is the key itself. So for primaryIndex=0, the secondaryIndex is the key. So it works. For primaryIndex=1, the keys are from 1000 to 1999. The secondaryIndex is key % 1000, which is key - 1000. So it is unique. Similarly for other primary indices. So actually, the solution correctly maps each key to a unique slot within the primary bucket. Therefore, the solution should be correct.

However, the space usage is high: it allocates arrays for each primary bucket that is touched. For example, if only keys in primaryIndex=0 are used, it allocates an array of size 1001. If keys in primaryIndex=1000 are used, it allocates an array of size 1000. The total number of primary buckets that might be used is 1001. So the worst-case space is 1001 * 1001 (for index0) + 1000 * 1000 (for other indices) = 1001*1001 + 1000*1000 = 1002001 + 1000000 = 2002001 integers. This is about 2e6 integers, which is 8e6 bytes (each integer is 4 bytes) which is about 8MB. This is acceptable in Java? But the problem says "at most 10^4 calls", meaning only 10^4 keys will be inserted. So the solution might allocate arrays for many primary buckets even if only one key in that bucket is used. For example, if we insert key=0, key=1000, key=2000, ... key=1000000, then we use 1001 primary buckets, and each primary bucket (except index0) has an array of size 1000, and index0 has size 1001. So total space is 1001 + 1000*1000 = 1001 + 1000000 = 1001001 integers. But we only have 1001 keys. So the space usage is 1000 times more than necessary.

In contrast, the reference solution uses linked lists and only allocates nodes for the keys actually stored. So the space is O(n) where n is the number of keys.

Also, the student's solution initializes each secondary array with -1 using Arrays.fill. This is O(1000) per primary bucket, which is acceptable but not necessary if we only set the used slots to -1? Actually, it is necessary because the get method must return -1 for unused slots. But the initialization is done when the primary bucket is first used. So if we have 10^4 keys spread across 10 primary buckets, we initialize 10 arrays each with 1000 or

VERDICT: NEEDS_IMPROVEMENT

